PR #2, opened 2017-04-05
example json file and hotloading working

schuyler1d:


src/actions/petitionActions.js
```diff
@@ -0,0 +1,46 @@
+export const actionTypes = {
+  'FETCH_PETITION_REQUEST': 'FETCH_PETITION_REQUEST',
+  'FETCH_PETITION_SUCCESS': 'FETCH_PETITION_SUCCESS',
+  'FETCH_PETITION_FAILURE': 'FETCH_PETITION_FAILURE'
+};
+
+let API_URI = process.env.API_URI;
```
mathemagica: should this be wrapped in some config mgmt class?

src/actions/petitionActions.js
```diff
@@ -0,0 +1,46 @@
...
+
+export function loadPetition(petition_slug) {
+  return function(dispatch) {
+    console.log('from mapDispatchToProps');
```
mathemagica: a few console logs left in

src/components/petition.js
```diff
@@ -97,7 +101,23 @@ Petition.propTypes = {
...
+    'foo': 1
+    //petition: state.petitions[ownProps.petitionkey]
+  }
+}
+
+function mapDispatchToProps(dispatch, ownProps) {
+  return {
+    fooBar: () => {
+      console.log('from mapDispatchToProps');
+      dispatch({type:'FETCH_PETITION_REQUEST', foo:1, bar:2})
```
mathemagica: is it worth it at this stage extracting action creators instead of dispatching action objects inline - could help prevent typos in event names (remember this coming up in SVG)

src/pages/sign-petition.js
```diff
@@ -1,36 +1,25 @@
...
+import {actions as petitionActions} from '../actions/petitionActions.js';
 
 class SignPetition extends React.Component {
   constructor (props) {
     super(props);
-    this.state = { petition: null };
   }
 
   componentWillMount () {
-    let urlKey = 'petitions/' + this.props.params.petition_slug;
```
mathemagica: i'm likely not understanding this well enough yet, but is componentWillMount supposed to explicitly call a dispatch method that could let other components know it's done mounting?

src/actions/petitionActions.js
schuyler1d: I really like this idea.  Let's follow up and add this.

src/actions/petitionActions.js
schuyler1d: going to leave those in without shame for a while -- they help me anyway :)

src/components/petition.js
schuyler1d: yes, this is dead/sample code, we need to delete, but the actual ones are exactly as you say.

src/pages/sign-petition.js
schuyler1d: i think react handles the negotiation of what subcomponents need to re-render, etc.

I'm also not completely certain this is the best place to run loadPetition() -- Scott's focus on figuring how what triggers a re-render or unnecessary renders at the beginning, I think is the right way to decide this.  More investigation needed here.

src/components/petition.js
mathemagica: ++

src/pages/sign-petition.js
mathemagica: makes sense. we'll keep learning more about this

src/components/petition.js
```diff
@@ -1,9 +1,13 @@
 import React from 'react';
+import {connect} from 'react-redux';
+import {bindActionCreators} from 'redux';
+
 import PetitionLoader from '../loaders/petition.js';
 import SignatureAddForm from './signature-add-form.js';
 
 class Petition extends React.Component {
   constructor (props) {
+    console.log(props);
```
mathemagica: this is a comment for line 15 / componentDidMount. instead of using a timeout, should this act like a reducer (or should it be composed together with a reducer) and wait for an event to be fired saying the petition data is ready, and then update its state with that new data?

src/pages/sign-petition.js
mathemagica: read a bit more about this. i think react components can subscribe to a particular redux store (the thing that collects up all the state and definitions of state changes), and by subscribing to a store they are able to then get notified when state changes happen, with redux's event bubbling that happens behind the scenes. one thing i'm not super sure about yet is: if all state changes need to happen within reducers, where does the API call go? should it be called not directly from the component but from inside a petition reducer?

src/actions/petitionActions.js
schuyler1d: fixed in 59b0c64

src/components/petition.js
sreynen: componentDidMount here is entirely example code of how this kind of thing (loading a component after the initial load) would work. It won't be needed here at all in production, but we don't yet have a better place to demonstrate this functionality, and we're not yet treating this as production code.

PR #6, opened 2017-04-10
remove mapDispatchToProps, and instead import  actions directly

sreynen:
I think this is functionally the same, with less code, but maybe I'm missing something?

### DIFF
--- a/src/components/signature-add-form.js
+++ b/src/components/signature-add-form.js
@@ -52,6 +52,7 @@ class SignatureAddForm extends React.Component {
   }
 
   submit(event) {
+      const {dispatch, petition} = this.props;
       if (this.formIsValid()) {
         var odsiSignature = {
           'person': {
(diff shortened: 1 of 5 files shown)


sreynen:
This also stops passing `defaultState` to `configureStore()`, since that throws an error and doesn't otherwise seem to do anything.

src/components/signature-add-form.js
```diff
@@ -78,7 +79,7 @@ class SignatureAddForm extends React.Component {
           'comments': this.state.comment
         };
         console.log('odsiSignature in addform submit()', odsiSignature);
-        this.props.actions.signPetition(odsiSignature, this.props.petition);
+        dispatch(petitionActions.signPetition(odsiSignature, petition));
```
schuyler1d: I don't think this will (or should) work, in the general (or even specific) case.
signPetition needs the method to dispatch (new) things (like SIGN_PETITION_SUCCESS after an ajax async call), it doesn't return the thing to dispatch.
We could do something like `petitionActions.signPetition(odsiSignature, petition, dispatch)` (and then modifying `actions/petitionActions.js` which needs a change anyway here)
I *do* like this better than the mapDispatchToProps thing but it seems like a peculiar pattern to always call with dispatch -- i guess it's sort of like sending a callback, so, maybe.

Also, weren't you interested in somehow specifically loading in only signPetition (rather than all actions) somehow?
Depending on the code it creates, I'm for that explicit declaration  -- in this form, I would search for `petitionActions.` in the codebase to see what actions are called.

schuyler1d:
:+1: love this (now that I understand it :-)

PR #7, opened 2017-04-10
Dev base path

sreynen:
This allows dev loading at e.g. `/#/sign/outkast` instead of `/build/#/sign/outkast`


schuyler1d:
works for me :+1:

Issue #20, opened 2017-10-23
Add lint to Travis test check


sreynen:
#15 makes the code free of lint errors, so we can now use Travis to flag PRs that introduce new lint errors.

PR #21, opened 2017-10-23
lint and test

sreynen:
For #20.

Note: this assumes #19, so that should be reviewed and merged before this is reviewed.

### DIFF
--- a/src/components/signature-add-form.js
+++ b/src/components/signature-add-form.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import { connect } from 'react-redux'
 
+import { text2paraJsx } from '../lib.js'
 import { actions as petitionActions } from '../actions/petitionActions.js'
 
 class SignatureAddForm extends React.Component {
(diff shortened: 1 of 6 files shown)


src/components/signature-add-form.js
```diff
@@ -499,13 +497,7 @@ class SignatureAddForm extends React.Component {
 
 SignatureAddForm.propTypes = {
   petition: PropTypes.object.isRequired,
-  dispatch: PropTypes.func,
-  text2paras: PropTypes.func
+  dispatch: PropTypes.func
```
schuyler1d: is/should dispatch be a property if you're removing the `connect()` stuff?  but that's extra

sreynen: Thanks for catching that. It was a mistake to remove connect. At the time, I didn't really understand what `connect()` does, just that it was referencing `mapStateToProps`, which wasn't doing anything. But it also adds `dispatch`, which is needed here, so I re-added `connect()`, just without `mapStateToProps`.

PR #36, opened 2017-11-06
some sign-petition tweaks and a userData redux state stub

schuyler1d:


.eslintrc.json
```diff
@@ -4,6 +4,7 @@
   "rules": {
     "semi": ["error", "never"],
     "comma-dangle": ["error", "never"],
+    "no-underscore-dangle": ["off"],
```
sreynen: I think I'm against this change. I like this rule in general. There are a few places where we can't avoid violating it, but ESLint has a way to mark those more specifically.

local/api/v1/petitions/95983.json
```diff
@@ -0,0 +1 @@
+outkast.json
```
sreynen: Was this intended to be the contents of outkast.json?

src/actions/petitionActions.js
```diff
@@ -60,11 +60,29 @@ export function signPetition(petitionSignature, petition) {
       signature: petitionSignature
     })
     // TODO: actually submit signature
-    dispatch({
-      type: actionTypes.PETITION_SIGNATURE_SUCCESS,
-      petition,
-      signature: petitionSignature
-    })
+    console.log('about to submit signature', petitionSignature, petition)
```
sreynen: console.log should be removed before merge.

src/actions/petitionActions.js
```diff
@@ -60,11 +60,29 @@ export function signPetition(petitionSignature, petition) {
...
+      }).then(completion, function(err) {
+        dispatch({
+          type: actionTypes.PETITION_SIGNATURE_FAILURE,
+          petition,
+          signature: petitionSignature,
+          error: err
+      })
+      })
+    } else {
+      completion()
```
sreynen: Why not call dispatch directly here? `completion` doesn't seem to be used anywhere else, so not sure why it's defined at all.

src/components/signature-list-item.js
```diff
@@ -6,7 +6,8 @@ import { formatDate } from '../lib.js'
 const SignatureListItem = ({ signature, number }) => {
   const { user } = signature
   const { city, state } = user
-  const location = (city === '') ? state : `${city}, ${state}`
+  console.log('citystate', city, state)
```
sreynen: console.log should be removed before merge.

src/components/signature-list-item.js
```diff
@@ -6,7 +6,8 @@ import { formatDate } from '../lib.js'
...
+  const location = ((!city) ? state : `${city}, ${state}`)
```
sreynen: Are the outer parentheses needed here? Or added for readability?

src/config.js
```diff
@@ -1,5 +1,6 @@
 export const Config = {
   API_URI: process.env.API_URI,
+  API_WRITABLE: process.env.API_WRITABLE,
```
sreynen: New environment variables should also be added to https://github.com/MoveOnOrg/mop-frontend/blob/master/docs/REFERENCE--environment_variables.md

src/reducers/index.js
```diff
@@ -69,8 +69,20 @@ function petitionReducer(state = initialState, action) {
   }
 }
 
+function userReducer(state = {}, action) {
+  switch (action.type) {
+    case petitionActionTypes.PETITION_SIGNATURE_SUBMIT:
+      console.log('userReducer on submit', action)
```
sreynen: console.log

src/reducers/index.js
```diff
@@ -69,8 +69,20 @@ function petitionReducer(state = initialState, action) {
   }
 }
 
+function userReducer(state = {}, action) {
```
sreynen: This reducer isn't doing anything yet. I think we should leave placeholder code out of master and only add after it's functional.

.eslintrc.json
schuyler1d: this goes specifically against HAL and OSDI json structure (with _`links` and `_embed`) -- we shouldn't need to add exception files every time we traverse or create a valid OSDI object.

local/api/v1/petitions/95983.json
schuyler1d: this is a symlink -- i guess this is how github/git represents that.

src/components/signature-list-item.js
schuyler1d: I add them for readability.  My personal preferred syntax for trinary logic is `((condition) ? value1 : value2)`  This makes it easy to see that it's trinary logic (even when/if multi-line breaking on `?` and `:`).  This isn't just a syntax change.  `city` can be `null` -- so it's important to condition on all false-y values.

src/actions/petitionActions.js
schuyler1d: it's in line 75 just above `}).then(completion, function(err) {`

.eslintrc.json
sreynen: I disagree, but I don't care enough to block the PR over it.

src/components/signature-list-item.js
sreynen: The syntax change is the part I'm concerned with. I don't have a preference on what syntax we use, but I do have a preference that we use the same syntax throughout the code. So if we're making this syntax change here, I ask that we also make it everywhere else.

schuyler1d:
@sreynen  besides handling your comments, I've pushed some more logic for handling/retrieving session state.  Could you take a look at this PR again.

Besides usual code review stuff two big questions:
1. I'm particularly interested in your thoughts about how I load session state 'on load' here:
https://github.com/MoveOnOrg/mop-frontend/pull/36/files#diff-3ca5e313cdf02368ce63986a36207c94R14
This seems to have the advantage of truly loading a single time (rather than in a component, which might load many times.  E.g. I experimented with adding it to components/wrapper.js which did work, but then ran every time a page loaded.  We can obviously save store.userStore state and make sure it loads once, but running the tests at all seem possibly unuseful.  OTOH, happy to move it back there (or somewhere else)

2. Take a look at: https://github.com/MoveOnOrg/mop-frontend/pull/36/files#diff-f05d9d6bbe554232ede2373d483b579f and confirm that's what you had in mind for an API, or let me know how/what I should be calling/expecting.

sreynen:
On the questions:

1. I don't have a clear idea yet for how this should work, but I think we'll want to load session on every initial page load regardless of path (including static pages that can't populate in preload), so the nav always shows the appropriate login/logout links. Maybe we could dispatch it in app.js before render happens?

2. That looks good to me. I imagine we'll find use cases for adding more data, but that seems like enough for now.

Issue #39, opened 2017-11-15
Create nav-only build option for static pages


sreynen:
Navigation needs to be dynamic even on static pages, because it shows different links depending on session state, e.g. a "logout" link only appears if you're logged in. So we'll need a separate build for a simpler React app that only renders the nav and does't interact with the rest of the page.

PR #40, opened 2017-11-15
Add nav-only build option

sreynen:
For #39, this does 3 things:

1. Adds (and documents) `APP_ENTRY` environment variable to override the default /src/app.js
2. Adds a new /src/nav-app/app.js that targets ID "header" instead of "root"
3. Adds a new /src/nav-app/routes.js that loads only Nav component for all paths.

### DIFF
--- a/src/nav-app/routes.js
+++ b/src/nav-app/routes.js
@@ -1,0 +1,18 @@
+import React from 'react'
+import { Router, Route, browserHistory, hashHistory } from 'react-router'
+
+import { Config } from '../config'
+import { loadSession } from '../actions/sessionActions'
+import SignPetition from '../pages/sign-petition'
+import ThanksPage from '../pages/thanks'
+import Nav from '../components/nav'
+
+const baseAppPath = process.env.BASE_APP_PATH || '/'
+
+export const appLocation = (Config.USE_HASH_BROWSING ? hashHistory : browserHistory)
+
+export const routes = (store) => (
+  <Router history={appLocation}>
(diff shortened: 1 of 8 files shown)


src/nav-app/routes.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import { Router, Route, browserHistory, hashHistory } from 'react-router'
+
+import { Config } from '../config'
+import { loadSession } from '../actions/sessionActions'
+import SignPetition from '../pages/sign-petition'
```
schuyler1d: are these necessary? (SignPetition and ThanksPage)

webpack.config.js
```diff
@@ -7,7 +7,7 @@ var APP_DIR = path.resolve(__dirname, 'src');
 
 var config = {
   entry: {
-    javascript: APP_DIR + '/app.js'
+    javascript: process.env.APP_ENTRY || APP_DIR + '/app.js'
```
schuyler1d: I feel like if we're hosting multiple 'apps' in a single config, then they should build to different places as well.  Perhaps we should have a `./src/apps/` directory with both our default or `petition` app, and another `nav`
and then instead of `APP_ENTRY` being a path, it would just go to `APP_DIR + 'apps/'+ APP_ENTRY + '.js'` or whatever with it defaulting to `petition` or something.

Then lower down in this file, instead of everything building to `main.js` it should build to APP_ENTRY+'-main.js'` 
or maybe the whole nav thing should be its own repo and then it could have its own webpack.config, etc.

Issue #27, opened 2017-10-27
Create route for search page


sreynen:
I.e. https://petitions.moveon.org/find/

Don't need to complete the whole page, just a core page component that we can fill out in other issues.

PR #41, opened 2017-11-15
Adds search page route and search component

shakalee14:
Hey @sreynen - you created several issues like this. Were you thinking something like this for the route creation issues?

### DIFF
--- a/src/pages/search.js
+++ b/src/pages/search.js
@@ -1,0 +1,20 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { connect } from 'react-redux'
+
+class SearchPage extends React.Component {
+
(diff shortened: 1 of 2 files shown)


sreynen:
My thinking was different routes are going to have different levels of complexity, and rather than try to map those out in advance, we could just see how far we can get in a first pass on each, and create follow-up issues for whatever can't fit in the first pass. So I think this is a fine for a first pass, and we should also capture in new issues what's left to do, or at least what a good next step would be. I'll do that on this one to show what I mean.

👍 on this PR.

shakalee14:
https://github.com/MoveOnOrg/mop-frontend/issues/27

Issue #25, opened 2017-10-27
Create route for home page


sreynen:
I.e. https://petitions.moveon.org/

Don't need to complete the whole page, just a core page component that we can fill out in other issues.

PR #48, opened 2017-11-16
Adds route for home page and simple component

shakalee14:
for #25

### DIFF
--- a/src/routes.js
+++ b/src/routes.js
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Router, Route, browserHistory, hashHistory } from 'react-router'
+import { IndexRoute, Route, Router, browserHistory, hashHistory } from 'react-router'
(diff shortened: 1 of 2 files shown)


src/routes.js
```diff
@@ -14,6 +15,7 @@ export const appLocation = (Config.USE_HASH_BROWSING ? hashHistory : browserHist
 
 export const routes = (store) => (
   <Router history={appLocation}>
+    <Route path='/' component={Home} />
```
sreynen: I think we could do this within the wrapper route, and remove Nav and Footer from the component, with `IndexRoute`. Example from Spoke: https://github.com/MoveOnOrg/Spoke/blob/master/src/routes.jsx#L27

shakalee14: @sreynen 👍

PR #60, opened 2017-11-17
Adds billboard component to home page

shakalee14:


shakalee14:
for #25

src/components/billboard.js
```diff
@@ -0,0 +1,33 @@
...
+
+  render() {
+    return (
+      <div className="container">
+        <div className="row">
+          <div className="span12 billboard-background bump-bottom-1 clearfix">
+            <div className="billboard-video text-center center clearfix">
+                <div className="billboard-content">
+                  <span className="billboard-video">
+                    <iframe width="500" height="380" src="//www.youtube.com/embed/oMHH7FL66fo" frameBorder="0" allowFullScreen></iframe>
```
schuyler1d: please make this a property.  Probably the property should be a whole component which will render the whole iframe.
This can be the default video/component.
Also, if you hardcode a youtube video link keep `https://` -- that shouldn't be protocol-neutral.

src/components/billboard.js
```diff
@@ -0,0 +1,33 @@
...
+                  </span>
+                </div>
+              <div className="billboard-text">
+                <h1 className="white big-title">People Powered Petitions</h1>
+                <p className="white size-medium-large">There are more than 8 million MoveOn members. Tap into our shared people power and create progressive change.</p>
+                <a href="/create_start.html?source=petitionshomepage" className="button button-xl background-moveon-bright-red" style={billboardButton}>
```
schuyler1d: maybe add a `// TODO` for making this a `<Link>` when we have a creation page.

src/components/billboard.js
shakalee14: i can make a video component

src/components/billboard.js
shakalee14: You can't make comments here but I can make one above

PR #61, opened 2017-11-17
37 revamp html

lady3bean:


local/index.html
```diff
@@ -1,9 +1,9 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
-    <title>Petitions</title>
+    <title>MoveOn Petitions</title>
+    <link rel='stylesheet' href='//petitions.moveon.org/moui/prod/css/moui.css?version=119' type='text/css' media='all' />
```
sreynen: I believe I added this, but looking at it now, I'm not sure why. The next line seems to be adding the same CSS file at a different URL.

sreynen:
This is failing automated tests, which are looking for the old HTML. I think the solution is to update the tests to look for the new HTML instead.

local/index.html
schuyler1d: If it's not too much trouble, can we make 'MoveOn' rendered as a template variable or something here?  If it's trouble, then let's save it for another time.

src/components/nav.js
```diff
@@ -1,29 +1,53 @@
...
+        <div className="mobile visible-phone">
+          <div className="moveon-masthead pull-left">
+            <a href="/">MoveOn.org</a>
+          </div>
+        </div>
+
+        <div className="pull-left top-icons hidden-phone">
+          <div className="pull-left span2 petitions-partner-logo bump-top-1 margin-right-2 hidden-phone"></div>
+          <a className="icon-link-narrow icon-start" href="http://petitions.moveon.org/create_start.html?source=topnav">Start a petition</a>
+          <a className="icon-link-narrow icon-managepetitions" href="http://petitions.moveon.org/dashboard.html?source=topnav">Scott's Dashboard</a>
```
shakalee14: Remove Scott's name

src/components/footer.js
```diff
@@ -1,42 +1,44 @@
...
+      <div className="span4">
+        <div className="footer">
+          <h1 className="moveon-masthead">
+            <a href="#">MoveOn.org&reg;</a>
+          </h1>
+        </div>
+        <div className="row">
+          <div className="span4">
+            <a className="icon-start icon-link-narrow" href="http://petitions.moveon.org/create_start.html?source=bnav">Start a Petition</a>
+            <a className="icon-link-narrow icon-managepetitions" href="http://petitions.moveon.org/dashboard.html">Scott's Dashboard</a>
```
shakalee14: Remove Scott's name

Issue #54, opened 2017-11-16
Add search form component, add to home page


sreynen:
Until search page is done, this should actually submit to the search URL. After search is in React, we can skip the extra page load.

PR #62, opened 2017-11-20
Search component home page

shakalee14:


shakalee14:
for #54

src/components/searchbar.js
```diff
@@ -0,0 +1,81 @@
...
+  render() {
+    return (
+      <div className="container">
+        <div className="row">
+          <div className="span7 control-group bump-top-1">
+            <form className="search" method="get" action="/find/">
+              <div className="search">
+                <input name="q" placeholder="Search Petitions" type="text" className="margin-right-1 " />
+                <select style={selectStyle} className="margin-right-1" name="state">
+                  <option value="">All States</option>
```
schuyler1d: Can we split this out and share the state list with petition sigining page?

src/components/searchbar.js
```diff
@@ -0,0 +1,81 @@
...
+}
+
+export default class SearchBar extends React.Component {
+
+  render() {
+    return (
+      <div className="container">
+        <div className="row">
+          <div className="span7 control-group bump-top-1">
+            <form className="search" method="get" action="/find/">
```
schuyler1d: can we add a `// TODO ` comment on handling the search within the react app

src/components/searchbar.js
```diff
@@ -0,0 +1,81 @@
+import React from 'react'
+
+const selectStyle = {
+  display: 'inline',
+  width: '100px'
+}
+
+export default class SearchBar extends React.Component {
```
schuyler1d: This is a shortcut, but if you are making a real component here, let's setup the file the 'standard' way and define the class up here, and then export at the bottom.  That way, if/when we need to add proptypes or other parts, we can just edit in-place.

Alternatively, you can start this out as 'just a function' (see e.g. the footer component).  However, since we'll need to handle the form at some point, it doesn't seem worth going back.

src/components/searchbar.js
schuyler1d: Actually, it looks like @lady3bean 's PR:
https://github.com/MoveOnOrg/mop-frontend/pull/61/files#diff-c645b827af8c79599f96b0024db4ea62
has a state-select, so ideally let's merge both and then combine

shakalee14: I saw @lady3bean's, We could merge that one and I can refactor

PR #66, opened 2017-11-22
refactor sig count component to be more testable, and subsequent tests

lady3bean:


test/components/singnature-list-item.js
```diff
@@ -0,0 +1,28 @@
...
+import { shallow } from 'enzyme';
+
+import SignatureListItem from '../../src/components/signature-list-item';
+
+describe('<SignatureListItem />', () => {
+
+	const user = {
+		city: 'Nottingham',
+		state: 'England',
+		name: 'Ada Lovelace',
```
shakalee14: 💯

Issue #34, opened 2017-11-02
Add session awareness to nav.js


sreynen:
When logged in, "EDIT ACCOUNT" and "LOGOUT" links should be added. Also, currently "MANAGE PETITIONS" link changes to "{firstname}'S DASHBOARD" on legacy site, but not sure whether or not new design intends to maintain that switch.

If we can't do this entirely client-side with the current login process (e.g. by reading a cookie), this issue should maybe be deferred until we have a new login process.

shakalee14:
i think this issue needs login & sign up routes first? but maybe not?

schuyler1d:
@shakalee14 https://github.com/MoveOnOrg/mop-frontend/pull/65 should provide session state which you can use in nav, as well.  (just testing `store.userStore.signonId` as long as you've hooked the component up to redux)

PR #68, opened 2017-11-27
Refactors nav to component, adds redux connector for userStore and co…

shakalee14:
…nditional for guest and user links

for #34


shakalee14:
@sreynen - is this the right track?

Issue #50, opened 2017-11-16
Create "hot petitions" component, add to home page


sreynen:
This should use API at /top-petitions.json

PR #72, opened 2017-11-30
hot petitions

sreynen:
For #50


src/components/petition-preview.js
```diff
@@ -0,0 +1,32 @@
...
+import { text2paraJsx } from '../lib.js'
+
+const PetitionPreview = ({ petition, source }) => {
+  const url = `/sign/${petition.name}?source=${source}`
+  const statement = text2paraJsx(petition.summary)
+  return (
+    <div>
+      <article>
+        <div>
+          <h4><a href={url} className="signthistitle">{petition.title}</a></h4>
```
schuyler1d: These should use routing `<Link>` if possible.
```
import {Link} from 'react-router';
...
<Link to={url} className="btn">Map</Link>
```
though I think you might need to play with that or the ?source won't work. (and you might need to remove the initial '/'.  If that doesn't work, try `onClick={(evt) => {appLocation.push(url)}}` with `import { appLocation } from '../routes.js'`

We need this or it will reload the whole react app on click rather than route within the react app.

src/reducers/index.js
```diff
@@ -86,6 +87,20 @@ function petitionReducer(state = initialPetitionState, action) {
...
+    case petitionActionTypes.FETCH_TOP_PETITIONS_SUCCESS:
+      const {petitions, pac, megapartner} = action
+      return Object.assign({}, state, {
+        petitions: petitions.reduce((addedPetitions, petition) => {
+          addedPetitions[petition.name] = petition
+          addedPetitions[petition.petition_id] = petition
+          return addedPetitions
+        }, state.petitions),
+        topPetitions: Object.assign({}, state.topPetitions, {
+          [pac]: Object.assign({}, state.topPetitions[pac], {
```
schuyler1d: instead of making this hierarchical, consider making it a single key, i.e. `${pac}_${megapartner}` -- this code will look more straightforward, and it makes it clear that unless you have the exact values for the search, then we'll need to requery it (I think, right?)

src/actions/petitionActions.js
```diff
@@ -53,6 +57,48 @@ export function loadPetition(petitionSlug) {
   }
 }
 
+export function loadTopPetitions(pac, megapartner) {
+  let urlKey = `top-petitions`
+  if (pac === 1) {
```
schuyler1d: should this just be true?
Also why not just treat `if (pac...)` outside the other loop of `if megapartner...`?
I think we are not supposed to show pac petitions on petitions.moveon.org (so pac=0 should occur, too)

sreynen: Changed pac to a simpler boolean check. I think "pac=0" is the default, so doesn't need to be passed. Or if it's not the default, we should change that on the server side, as it is the default in the legacy code.

schuyler1d: The legacy `Petition.pm::top_petitions` does default to pac=0.
At least currently, I don't default the filter in mop's api call:
https://github.com/MoveOnOrg/mop/pull/38/files#diff-0c4ecc15122dee906d52d84b63d629feR199
We could make it the default there, but it seemed like we might want to be able to query the total top-petitions list and if it defaults to 0 we'd need a separate variable.  But lmk, and I can change `mop`

PR #73, opened 2017-11-30
thanks page (more work)

schuyler1d:
Figured I should do a push and a code-review before continuing. This implements:

* shareclick tracking, 
* sharebandit support, 
* fixes sign-page html a bit (class => className, etc)
* thanks redirect in actions instead of page, to fix back button

Still todo (for future PRs):
* google analytics 'conversion' tracking
* short code/urls (for the anonymous signature use-case)
* next petition

### DIFF
--- a/src/components/signature-add-form.js
+++ b/src/components/signature-add-form.js
@@ -86,7 +86,7 @@ class SignatureAddForm extends React.Component {
         },
         comments: this.state.comment
       }
-      dispatch(petitionActions.signPetition(osdiSignature, petition, /*redirectOnSuccess*/true))
+      dispatch(petitionActions.signPetition(osdiSignature, petition, {redirectOnSuccess: true}))
     }
     event.preventDefault()
     return false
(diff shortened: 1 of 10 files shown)


mathemagica:
👍

src/components/signature-add-form.js
```diff
@@ -86,7 +86,7 @@ class SignatureAddForm extends React.Component {
         },
         comments: this.state.comment
       }
-      dispatch(petitionActions.signPetition(osdiSignature, petition))
+      dispatch(petitionActions.signPetition(osdiSignature, petition, /*redirectOnSuccess*/true))
```
sreynen: If this parameter seems worth naming, maybe we should switch to an options object so we can actually do that, like `dispatch(petitionActions.signPetition(osdiSignature, petition, {redirectOnSuccess: true}))`

schuyler1d: good idea.  I like that pattern

PR #76, opened 2017-12-01
fix path glob for eslint, so it checks subdirectories

sreynen:
For #75


sreynen:
Note: this fails tests because it doesn't yet fix all the lint errors we've introduced while this bug has existed.

PR #79, opened 2017-12-04
sign page header stuff mostly done. 

schuyler1d:
TODO: operational flag form, abstract out the "MoveOn"s and/or urls (or not)
also everything below


mathemagica:
The possibly_out_of_date tag messaging works, as does the "status": "Bad" messaging. It looks like "MoveOn has not yet reviewed this petition. If you agree with it, please sign it!" always shows up no matter what though - is this supposed to be controlled with a config flag too?

Also, I'm not logged in and view http://0.0.0.0:8080/#/sign/outkast-opposite and I see "UNDEFINED'S DASHBOARD" up in the header. Should this not display at all if I'm not logged in?

screen shot 2017-12-07 at 11 14 32 am

schuyler1d:
It needs to have status: "Verified" to not have that message, as here:
https://github.com/MoveOnOrg/mop-frontend/pull/79/files#diff-3d9c6c90e33cee2a80587fda2c505e81R24

I can debug the header if you like, but this PR is not related to the header.

mathemagica:
Thanks for the clarifications. 👍

Issue #80, opened 2017-12-04
Fix Search Component Formatting Issue on Home Page


shakalee14:
- Make it look like current search section on petitions.moveon.org

PR #100, opened 2017-12-05
Refactors state select to take text and style

shakalee14:
for #80 
- note depending on network speed, proper google font may not load which will make the font size/type look slightly different
- most styling comes from static css
- the styling added was to allow us to use state-select logic in different places, but with different styles
- the text refactor was added because petitions.moveon.org state select and sign petition pages on live site have different text in the selector.

### DIFF
--- a/src/components/form/state-select.js
+++ b/src/components/form/state-select.js
@@ -81,7 +81,8 @@ const StateSelect = ({ onChange, selectText, style }) => (
 
 StateSelect.propTypes = {
   onChange: PropTypes.func,
-  style: PropTypes.object
+  style: PropTypes.object,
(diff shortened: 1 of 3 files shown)


src/components/form/state-select.js
```diff
@@ -78,7 +80,9 @@ const StateSelect = ({ onChange }) => (
 )
 
 StateSelect.propTypes = {
-  onChange: PropTypes.func
+  onChange: PropTypes.func,
+  style: PropTypes.object,
+  selectText: PropTypes.object
```
sreynen: Shouldn't this be `PropTypes.string`? Looks like the values are strings, not objects.

Issue #92, opened 2017-12-04
Update title and meta description for home page


sreynen:
```
<title>MoveOn Petitions - Sign or Start a Petition</title>
<meta name=”description” content=”Sign people-powered petitions from across the country, start your own, and view past petition victories.”>
```

shakalee14:
Will close this since google reliably uses server-rendered meta tags

PR #101, opened 2017-12-05
Adds meta name and changes title

shakalee14:
for #92

### DIFF
--- a/local/index.html
+++ b/local/index.html
@@ -1,9 +1,10 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
-    <title>MoveOn Petitions</title>
+    <title>MoveOn Petitions - Sign or Start a Petition</title>
+    <meta name="description" content="Sign people-powered petitions from across the country, start your own, and view past petition victories.">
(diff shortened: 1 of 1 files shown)


sreynen:
I think we need to add these to both the server side, for the initial page load, and the React app, for pages discovered via JS crawling, which is [apparently a thing](https://searchengineland.com/tested-googlebot-crawls-javascript-heres-learned-220157). But to Sky's point, neither of those will be using files from /local, since that's only used in dev testing. So agree this is fine, but also that it doesn't solve #92.

shakalee14:
@sreynen - Maybe I'm not understanding this issue, but in my understanding `build/index.html` is created via `local/index.html` after running `npm run build`, and then that's what the app will serve. So it is used because it creates `build/index.html`

sreynen:
I believe `build/index.html` is what `npm run dev` will serve, though I'm not entirely sure about that. I am sure that the live site doesn't use that file at all. The initial HTML for any request is served by Django (e.g. https://github.com/MoveOnOrg/mop/blob/master/petitions/urls.py#L32) and everything after that is rendered by React JS. So for mop-frontend, we need to make the React code update the title and meta tags. And the Django side has a separate issue, https://github.com/MoveOnOrg/mop/issues/46

Issue #83, opened 2017-12-04
Hot petitions teaser text is too long


sreynen:
Legacy site seems to truncate with "..." at some point, and new site should match. Currently it outputs the entire summary.

PR #102, opened 2017-12-05
adds ellipsize function to lib

lady3bean:


lady3bean:
Fixes #83

lady3bean:
yeah, in the end it turned out to be a good a nicely flexible solution! think it was a better strategy than the css, though it did make sense for us to consider doing it through styling first

Issue #95, opened 2017-12-04
When logged out, get "undefined's Dashboard" in nav


sreynen:
Session API is returning `{"identifiers": []}`, which should show the guest nav, but it's showing a broken version of authenticated nav instead.

PR #103, opened 2017-12-05
update session checks for nav, add reusable navlink component, and update links

sreynen:
For #95, but also found in the process some of the links didn't match the live site, and since all the links have the same markup, moved that to a separate component to ease future restyling.

### DIFF
--- a/src/components/nav.js
+++ b/src/components/nav.js
@@ -64,7 +64,7 @@ const Nav = ({ user }) => {
             <span className='icon-th-list'></span>
           </a>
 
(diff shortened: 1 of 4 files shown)


src/components/nav.js
```diff
@@ -62,7 +64,7 @@ const Nav = ({ user }) => {
             <span className='icon-th-list'></span>
           </a>
 
-          {user ? userLinks : guestLinks}
+          {user.identifiers && user.identifiers.length ? userLinks : guestLinks}
```
schuyler1d: If this is userStore you can test on user.sigonId here.

The state doesn't have to only be the osdi fields

PR #105, opened 2017-12-06
Travis deploy

sreynen:
This will auto-deploy changes in master branch to mop-static-stage S3 bucket. This is a step we do for end-to-end testing pretty often, and this would remove that step.


sreynen:
Yeah, this is stage-only at this point. We might want to look at doing something similar for prod down the road, but not now.

PR #108, opened 2017-12-07
rework sign page for reliable validation

schuyler1d:
Some notes:
* I couldn't get stateValidationError in state-or-region-input.js to update based on the parent component's state.  I gave up and am printing it out within the signature-add-form (parent) component, but I'd appreciate the right pattern for this (the existing one wasn't updating), or we should drop it from the sub-component
* I had to move the `required` dict into state because it was occasionally updated by state (e.g. for volunteering, then we require a phone)


mathemagica:
tested this locally and validation errors work so generally 👍  . taking a wild guess here on what the right pattern should be, but did you try wrapping the setup of stateValidationError in state-or-region-input within a componentDidMount ?

schuyler1d:
@mathemagica
Can you say more about componentdidmount? Maybe we could chat tomorrow

PR #116, opened 2017-12-11
tests for optin warnings/checkboxes and clearing out bugs too

schuyler1d:


schuyler1d:
Two interesting tidbits I learned during this PR:
1. You can add multiple attributes to css query requirements like this: `input[name="thirdparty_optin"][type="checkbox"]`
2. Sometimes, it's useful adding a custom message with a failure, using the second argument of `expect()` for chai, e.g. https://github.com/MoveOnOrg/mop-frontend/pull/116/files#diff-e1018fbf6177cbaaf1c24d27decec211R139

PR #117, opened 2017-12-11
add partner nav logo when appropriate on partner homepage and …

### LINKED ISSUE #69: Add page and components for megapartner home pages
Should be able to load /:partner-name URLs and get generally the same content seen on the legacy site, e.g. http://petitions.moveon.org/americasvoice/ For the context of this issue, it doesn't need to be an exact match (though that would be great), just close enough that we can easily note the remaining work in more specific issues.

Most of this content is already in the home page, but needs to send the megapartner param to the API in this context.

schuyler1d:
… petitions

* Fixes #69 (Add nav logo/url for partners) on petitions/top-petitions page
* tracks google conversions on thanks event
* Adds featured_image_url (an osdi field) to support embedded petition->image (probably not used anymore, but we implemented it :-)

### DIFF
--- a/src/components/nav.js
+++ b/src/components/nav.js
@@ -4,7 +4,7 @@ import { connect } from 'react-redux'
 
 import NavLink from './nav-link'
 
-const Nav = ({ user }) => {
+const Nav = ({ user, nav }) => {
   const userLinks = (
     <div className='pull-right bump-top-1 span-7 top-menu'>
       <ul className='nav collapse nav-collapse'>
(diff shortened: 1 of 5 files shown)


src/components/nav.js
```diff
@@ -28,13 +28,18 @@ const Nav = ({ user }) => {
   )
 
   const userDashboardLink = (
-    <a className='icon-link-narrow icon-managepetitions' href='http://petitions.moveon.org/dashboard.html?source=topnav'> {`${user.given_name}'s`} Dashboard</a>
+    <a className='icon-link-narrow icon-managepetitions' href='http://petitions.moveon.org/dashboard.html?source=topnav'> {`${user.given_name}&#39;s`} Dashboard</a>
```
shakalee14: nice catch

PR #118, opened 2017-12-12
thanks-next-petition suggestion component. with some changes to the loadTopPetitions action

schuyler1d:
random reflection: This semi-dodges the 'pop an item to a queue' challenge.  We depend on petitionStore.nextPetitions getting filtered out by things that are marked within state or frrom the api as signed.  Then it's just the next in that queue.

It also depends on backend changes in https://github.com/MoveOnOrg/mop/pull/58

### DIFF
--- a/src/actions/petitionActions.js
+++ b/src/actions/petitionActions.js
@@ -78,7 +78,10 @@ export function loadTopPetitions(pac, megapartner, forceReload) {
       topPetitionsKey
     })
     const { userStore, petitionStore } = getState()
-    if (!forceReload && petitionStore.topPetitions[topPetitionsKey]) {
+    if (!forceReload
+        && petitionStore
+        && petitionStore.topPetitions
+        && petitionStore.topPetitions[topPetitionsKey]) {
       return dispatch({
         type: actionTypes.FETCH_TOP_PETITIONS_SUCCESS,
         useCache: true,
@@ -92,7 +95,7 @@ export function loadTopPetitions(pac, megapartner, forceReload) {
     if (megapartner) {
       query.push(`megapartner=${megapartner}`)
     }
-    if (userStore.signonId) {
+    if (userStore && userStore.signonId) {
       query.push(`user=${userStore.signonId}`)
     }
     const queryString = ((query.length) ? `?${query.join('&')}` : '')
--- a/src/actions/petitionActions.js
+++ b/src/actions/petitionActions.js
@@ -38,7 +38,10 @@ export function loadPetition(petitionSlug, forceReload) {
       slug: petitionSlug
     })
     const { petitionStore } = getState()
-    if (!forceReload && petitionStore.petitions[petitionSlug]) {
+    if (!forceReload
+        && petitionStore
+        && petitionStore.petitions
+        && petitionStore.petitions[petitionSlug]) {
       return dispatch({
         type: actionTypes.FETCH_PETITION_SUCCESS,
         petition: petitionStore.petitions[petitionSlug],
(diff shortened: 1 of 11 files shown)


src/actions/petitionActions.js
```diff
@@ -57,40 +68,51 @@ export function loadPetition(petitionSlug) {
...
-    } else {
-      urlKey += '?pac=1'
-    }
-  } else if (megapartner) {
-    urlKey += `?megapartner=${megapartner}`
-  }
-
-  return (dispatch) => {
+export function loadTopPetitions(pac, megapartner, forceReload) {
+  // topPetitionsKey must not just be truthily equal but exact
```
sreynen: This feels like we're working around a problem that would be better fixed, i.e. if we don't send the wrong types to this function then we won't need to do type correction.

src/components/thanks-next-petition.js
```diff
@@ -0,0 +1,115 @@
...
+      secondsLeft: this.startSeconds + 1,
+      intervalListener: null
+    }
+    this.startCountdown = this.startCountdown.bind(this)
+    this.tickCounter = this.tickCounter.bind(this)
+  }
+
+  componentWillMount() {
+    if (!this.props.nextPetitionsLoaded) {
+      this.props.dispatch(loadTopPetitions(this.props.entity === 'pac'))
```
sreynen: Suggest making this explicitly send the full parameters with expected types, i.e. `loadTopPetitions(this.props.entity === 'pac' ? 1 : 0, '', false)`

src/components/thanks-next-petition.js
```diff
@@ -0,0 +1,115 @@
...
+    }
+    const newSecondsLeft = this.state.secondsLeft - 1
+    this.setState({ secondsLeft: newSecondsLeft })
+    if (newSecondsLeft <= 0) {
+      this.redirectToNext()
+    }
+    return
+  }
+
+  nextUrl() {
```
sreynen: Looks like this is only ever called in `redirectToNext()`, and that has to repeat some of the logic used here, so would be simpler to just move the logic there.

src/actions/petitionActions.js
schuyler1d: I think of it as defensive coding.  We could put an assert here instead, but it seems just as easy to just make sure our key is the same format for all situations.

src/components/thanks-next-petition.js
```diff
@@ -0,0 +1,115 @@
...
+            : '')
+  }
+
+  redirectToNext() {
+    if (this.state.intervalListener) {
+      clearInterval(this.state.intervalListener)
+      this.setState({ intervalListener: null })
+    }
+    const nextUrl = this.nextUrl()
+    if (nextUrl) {
```
sreynen: Per the comment above, this could be `if (nextPetition) { appLocation.push(`/sign/${nextPetition.name}?source=share_chain`) }` and skip the entire `nextUrl` abstraction.

src/components/top-petitions.js
```diff
@@ -42,7 +42,9 @@ TopPetitions.propTypes = {
 const mapStateToProps = (store, ownProps) => {
   const { pac, megapartner } = ownProps
   const topPetitionsState = store.petitionStore.topPetitions
-  const topPetitionsKey = `${pac}--${megapartner}`
+  // topPetitionsKey must not just be truthily equal but exact
```
sreynen: Just above this, `TopPetitions.propTypes` says pac must be a number and megapartner must be a string. If that's true, this conversion is unnecessary. If that's not true, we should change those propTypes, but I'd vote to make it true.

src/components/thanks-next-petition.js
schuyler1d: this is so `nextUrl()` can be tested

schuyler1d:
closing PR, since I just merged the reviewed commits (with a tweak to address review) and then accidentally pushed another commit here.

Issue #113, opened 2017-12-11
Add more tests to signature-add-form


schuyler1d:
Look for `TODO` in `test/components/signature-add-form.js` for a start, but adding additional tests for other scenarios are good, too!

lady3bean:
this still needs more tests

PR #126, opened 2017-12-14
test for updating state requirement for non-US addresses

lady3bean:
issue #113


test/components/signature-add-form.js
```diff
@@ -204,9 +204,20 @@ describe('<SignatureAddForm />', () => {
...
-    // it('TODO:non-US address', () => {})
     // it('TODO:logout/unrecognize shows anonymous field list', () => {})
 
+    it('adding a non-US address updates requirements to not require state', () => {
+      const store = createMockStore(storeAnonymous)
+      const context = mount(<SignatureAddForm {...propsProfileBase} store={store}/>)
+      const component = unwrapReduxComponent(context)
+      const reqFieldsBefore = component.updateRequiredFields(true)
+      expect(typeof reqFieldsBefore.state).to.be.equal('string')
+      component.setState({address1:'123 main', city:'Sydney', country:'Australia', name:'John Smith', email:'hi@example.com', zip:'60024'})
```
schuyler1d: If you're going to use an Australian use-case, might as well do the 4-digit postcode, too:
https://en.wikipedia.org/wiki/Postcodes_in_Australia#Format

test/components/signature-add-form.js
```diff
@@ -204,9 +204,20 @@ describe('<SignatureAddForm />', () => {
...
+      const reqFieldsAfter = component.updateRequiredFields(true)
+      expect(typeof reqFieldsAfter.state).to.be.equal('undefined')
+      expect(component.formIsValid()).to.be.equal(true)
+    })
```
schuyler1d: can we add a test to confirm that the data is in `component.getOsdiSignature()`?  At least the state/region, country and postcode.

test/components/signature-add-form.js
lady3bean: interesting - it fails with a 4 digit zip code

Issue #43, opened 2017-11-15
Make search route respond to query parameters


sreynen:
Current handled parameters are: q, state, and page, maybe more. E.g.: https://petitions.moveon.org/find/?q=taxes&state=CO&page=2

This is currently blocked on https://github.com/MoveOnOrg/mop/issues/48

See also #42.

shakalee14:
Even though this is attached to https://github.com/MoveOnOrg/mop/pull/56, which is why it's blocked, I can still get started on this issue.

Issue #42, opened 2017-11-15
Add default (no search param) content to search page


sreynen:
Should match what's currently on https://petitions.moveon.org/find/

This is currently blocked on https://github.com/MoveOnOrg/mop/issues/48

PR #133, opened 2017-12-22
42 default content search page

shakalee14:
This isn't ready, but I'm leaving notes.

CURRENTLY WORKING:
if you have the mop server running, visit : http://localhost:8000/api/v1/search/petitions.json
if that returns petitions, then visit: http://localhost:8080/#/find . That url should return petitions in SearchResults component. Then visit: http://localhost:8080/#/find?query=cats . That should return petitions related to cats.

**PREVIOUS TODOS:**
- make search bar component post action to correct url [x]
- search results pagination component: click on page and it adds to end of query url [x]
- when you click on a page # under petition results, it should go to that page # in the url [x]
- make sure previous search results are cleared out before new search results [x]
- search results pagination component: fix highlight of current page [x]
- search bar component: when state is selected , add to url [x]
- double check pagination works server side : https://github.com/MoveOnOrg/mop/issues/87 [x] 
- make sure page options increase (or decrease) as you click on higher or lower pages (see example:
http://petitions.moveon.org/find/?page=5&) [x]

**LAST TWO TODOS:**
- add tests [ x ]
- With search result pages at the bottom of the page, (rendered via search-result-pagination.js). I use <Link /> components from react-router. This will correctly update the route but doesn't return the correct results yet (only when you refresh) --> this should be connected to searchPetitions() probably [ x ]
  
**What this PR does:**
Note: need to be connected to back end and running mop server to test PR
- Correctly renders search results based on input from search bars on home page and search page 
- Returns default search results when you visit `0.0.0.0:8080/#/find`
- Returns correct results when you click on page numbers (on bottom of find page)


shakalee14:
This PR will handle #43  and #42

shakalee14:
- [ ] With search result pages at the bottom of the page, (rendered via search-result-pagination.js). I use <Link /> components from react-router. This will correctly update the route but doesn't return the correct results yet (only when you refresh) --> this should be connected to searchPetitions() probably
-[ ] add tests

src/components/searchbar.js
```diff
@@ -1,27 +1,113 @@
...
+  }
+
+  submitQuery(e) {
+    e.preventDefault()
+    const dispatch = this.props.dispatch
+
+    const query = this.state.query
+    const selState = this.state.selectState || ''
+
+    if (query !== '' && selState === '') {
```
schuyler1d: This looks correct, but it's a little difficult to parse all the if branches, and if there were an error, it would be difficult to spot.  I might have done this with something like:
```
let queryString = []
if (this.state.query) {
   queryString.push(`q=${this.state.query}`)
}
if (this.state.selectState) {
  queryString.push(`state=${this.state.selectState}`)
}
if (queryString.length) {
  const fullQuery = queryString.join('&')
  this.props.router.push(`/find/?${fullQuery}`)
}
```

src/components/searchbar.js
```diff
@@ -1,27 +1,113 @@
...
+      this.props.router.push(`/find/?q=${this.state.query}`)
+    } else if (query !== '' && selState === '') {
+      this.props.router.push(`${this.state.query}`)
```
schuyler1d: Is this intended without the /find/?q= ?  If so, can you add a comment as to why?  If not, then maybe my version above makes this simpler.

src/components/searchbar.js
```diff
@@ -1,27 +1,113 @@
...
+    return (
+      <div>
+        {size === 'long' ? longSearchBar : shortSearchBar}
+      </div>
+    )
+  }
+}
+
+SearchBar.propTypes = {
+  size: PropTypes.string,
```
schuyler1d: since this just switches on 'long' maybe we should make it a boolean `isLong` instead, and the above state can decide.

src/components/searchbar.js
```diff
@@ -1,27 +1,113 @@
...
+  router: PropTypes.object,
+  query: PropTypes.string,
+  dispatch: PropTypes.func,
+  selectState: PropTypes.string
+}
 
-export default SearchBar
+export default connect()(withRouter(SearchBar))
```
schuyler1d: I didn't know about this pattern -- and this might be the better one.  In petitionActions, I do
`import { appLocation } from '../routes.js'` and then `appLocation.push(...)`.
It seems like the plus of this, in the component, anyway, is that the 'component' "knows" that it might redirect stuff, and maybe it does that better with redux's structure.
OTOH, if we didn't have this, then the component could be a 'regular' component without any redux-itude (for easier testing, at least)

src/components/search-result-pagination.js
```diff
@@ -0,0 +1,52 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Link } from 'react-router'
+import queryString from 'query-string'
```
schuyler1d: I didn't see this, but this is all parsed already with react's location. at
`this.props.location.query`  (see e.g. src/pages/thanks.js)

src/components/search-result-pagination.js
```diff
@@ -0,0 +1,53 @@
...
+import PropTypes from 'prop-types'
+import { Link } from 'react-router'
+
+const SearchResultPagination = ({ resultCount, pageSize, currentPage, searchNavLinks, query }) => {
+
+  const q = query || ''
+  const nextPage = searchNavLinks ? searchNavLinks.next : ''
+  const pageNow = searchNavLinks ? searchNavLinks.url : ''
+
+  const parsedPreviousPageLink = (location.query !== undefined ? location.query.pageNow - 1  : 0)
```
mathemagica: realizing that we shouldnt ever parse these params out of location - the top-level search component parses stuff out from location and passes these values down to the child components. so even in the event of a hard page load, search-result-pagination.js should get these values from props

package.json
```diff
@@ -48,6 +48,7 @@
     "jsdom-global": "3.0.2",
     "mocha": "~3.2.0",
     "nock": "^9.0.13",
+    "query-string": "^5.0.1",
```
schuyler1d: now that you're not parsing the query string, you don't need this anymore.
Also, please remove the `package-lock.json` file from the diff.  After you do so, please rebase, so it doesn't delete someone's local file when you do so.

src/pages/thanks.js
```diff
@@ -27,7 +27,7 @@ class ThanksPage extends React.Component {
     return (
       <div>
         {(this.Thanks && this.props.petition ?
-          <this.Thanks petition={this.props.petition} user={this.props.user} /> :
+          <this.Thanks searchQuery={this.props.searchQuery} user={this.props.user} /> :
```
schuyler1d: why are you editing the thanks.js file?  I think this is all in the wrong place?

test/components/signature-add-form.js
```diff
@@ -86,15 +86,15 @@ describe('<SignatureAddForm />', () => {
       expect(context.find('input[name="address1"]').length).to.equal(1)
       expect(context.find('input[name="address2"]').length).to.equal(1)
       expect(context.find('input[name="city"]').length).to.equal(1)
-      // not testing state because state is a sub component
-      // expect(context.find('input[name="state"]').length).to.equal(1);
+      // Not testing state because state is a sub component
+      // Expect(context.find('input[name="state"]').length).to.equal(1);
```
schuyler1d: Please undo the auto-upper-case comments -- this isn't relevant to this PR and some were just annoying, but this is commented-out code that won't be correct anymore.  same below for `It(...)`

package.json
mathemagica: i think this is being used in searchbar.js still

src/components/searchbar.js
mathemagica: are you suggested we switch this to the other pattern, or stick with this pattern?

mathemagica: actually, appLocation seems easier to read / debug. making this change

src/components/search-results.js
```diff
@@ -0,0 +1,71 @@
...
+    const resultCount = searchResults.count
+    const pageSize = searchResults.page_size
+    const resultsEmbed = searchResults._embed
+    const searchNavLinks = searchResults._links
+
+    const resultsList = resultsEmbed.map((result, index) => (
+      <div className='result' key={index}>
+        <div className='result-text'>
+          <p className='result-name'>
+            <a className='size-medium-large font-heavy' href={`http://pac.petitions.moveon.org/sign/${result.short_name}/?source=search`}>
```
schuyler1d: remove the hard-coded url.  If this is osdi-ish result, then you can do `result._links.url`
Don't leave it, but if you wanted to, then you'd still need to worry about the entity and it should be `https`, not http

src/components/search-results.js
```diff
@@ -0,0 +1,71 @@
...
+                {result.name}
+            </a>
+          </p>
+          <p className='size-small'>http://pac.petitions.moveon.org/sign/{result.short_name}/</p>
```
schuyler1d: ditto here.

src/components/signature-add-form.js
```diff
@@ -204,7 +204,7 @@ class SignatureAddForm extends React.Component {
     const petitionBy = creator.name + (creator.organization
                                        ? `, ${creator.organization}`
                                        : '')
-    const iframeEmbedText = `<iframe src="http://petitions.moveon.org/embed/widget.html?v=3&amp;name=${petition.slug}" class="moveon-petition" id="petition-embed" width="300px" height="500px"></iframe>` // text to be copy/pasted
+    const iframeEmbedText = `<iframe src="http://petitions.moveon.org/embed/widget.html?v=3&amp;name=${petition.slug}" class="moveon-petition" id="petition-embed" width="300px" height="500px"></iframe>` // Text to be copy/pasted
```
schuyler1d: If you're going to lint a line, then I'll judge the line -- change this to `https`

src/pages/thanks.js
```diff
@@ -37,16 +37,16 @@ class ThanksPage extends React.Component {
 }
 
 ThanksPage.propTypes = {
-  petition: PropTypes.object,
+  searchQuery: PropTypes.object,
```
schuyler1d: still please undo this -- thanks.js should be unchanged. and specificlaly shouldn't have a `searchQuery` property

src/components/search-result-pagination.js
```diff
@@ -0,0 +1,87 @@
...
+  changePage(newPage, newUrl) {
+    if (newPage !== this.props.currentPage) {
+      const dispatch = this.props.dispatch
+      dispatch(petitionActions.searchPetitions(this.props.query, newPage, this.props.selectState))
+      appLocation.push(newUrl)
+    }
+  }
+
+
+  debugger
```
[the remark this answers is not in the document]
mathemagica: derp

src/components/search-results.js
mathemagica: 👍

src/pages/thanks.js
mathemagica: now i have a paranoid theory linting did this

PR #134, opened 2017-12-27
more stuff for frontend to send signatures

schuyler1d:
* show_optin in osdi so we know whether they opted OUT

### DIFF
--- a/src/components/signature-add-form.js
+++ b/src/components/signature-add-form.js
@@ -50,6 +50,7 @@ class SignatureAddForm extends React.Component {
       petition: {
         name: petition.name,
         petition_id: petition.petition_id,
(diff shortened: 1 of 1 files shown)


src/components/signature-add-form.js
```diff
@@ -50,6 +50,7 @@ class SignatureAddForm extends React.Component {
       petition: {
         name: petition.name,
         petition_id: petition.petition_id,
+        show_optin: this.props.showOptinCheckbox,
```
sreynen: As this function is already destructuring `this.props` on line 48, suggest adding `showOptinCheckbox` there rather than referencing in within `this.props`, so we're not mixing different ways of pulling data out of `this.props` within the same function.

PR #135, opened 2018-01-08
make our API_URI do what our docs say, 

schuyler1d:
and also add a API_SIGN_PETITION override that can go anywhere


schuyler1d:
adding a couple of things to this while I am testing signing:
* the signature-add-form.js wasn't saving the comment and sending it to the backend, this fixes that.

PR #144, opened 2018-01-11
fix apostrophe in dashboard link

### LINKED ISSUE #142: Fix &#39; in dashboard link
E.g. mine says "SCOTT&#39;S DASHBOARD", should say "SCOTT'S DASHBOARD".

sreynen:
Fixes #142

### DIFF
--- a/src/components/nav.js
+++ b/src/components/nav.js
@@ -30,7 +30,7 @@ const Nav = ({ user, nav, organization }) => {
   )
 
   const userDashboardLink = (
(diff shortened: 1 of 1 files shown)


sreynen:
I think the HTML entity doesn't work because React encodes output automatically, so it gets double encoded. Changed it from straight to curly apostrophe. Does that work better?

schuyler1d:
Yes, thanks! lsquo is my new favorite character.

PR #156, opened 2018-01-18
fix user session info

schuyler1d:
After implementing the backend, this resolves some bugs/issues with the original implementation of session api and user state.

* We previously separated cookie loading vs. token loading.  Since a cookie doesn't actually indicate authenticated state, we need to combine them.
* Combining them makes the `TOKEN_SESSION_*` and `USER_SESSION_*` a bit redundant so we collapse them
* signature-add-form assumes if someone is identified then there's a `user.token` but we never implemented that (useful) simplification.  This does that based on parameters from the location and/or the session api data.

### DIFF
--- a/src/actions/sessionActions.js
+++ b/src/actions/sessionActions.js
@@ -6,9 +6,7 @@ export const actionTypes = {
   ANONYMOUS_SESSION_START: 'ANONYMOUS_SESSION_START',
   UNRECOGNIZE_USER_SESSION: 'UNRECOGNIZE_USER_SESSION',
   USER_SESSION_START: 'USER_SESSION_START',
-  USER_SESSION_FAILURE: 'USER_SESSION_FAILURE',
-  TOKEN_SESSION_START: 'TOKEN_SESSION_START',
-  TOKEN_SESSION_FAILURE: 'TOKEN_SESSION_FAILURE'
+  USER_SESSION_FAILURE: 'USER_SESSION_FAILURE'
 }
 
 export function unRecognize() {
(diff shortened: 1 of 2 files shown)


src/actions/sessionActions.js
```diff
@@ -6,9 +6,7 @@ export const actionTypes = {
   ANONYMOUS_SESSION_START: 'ANONYMOUS_SESSION_START',
   UNRECOGNIZE_USER_SESSION: 'UNRECOGNIZE_USER_SESSION',
   USER_SESSION_START: 'USER_SESSION_START',
-  USER_SESSION_FAILURE: 'USER_SESSION_FAILURE',
-  TOKEN_SESSION_START: 'TOKEN_SESSION_START',
-  TOKEN_SESSION_FAILURE: 'TOKEN_SESSION_FAILURE'
+  USER_SESSION_FAILURE: 'USER_SESSION_FAILURE'
```
schuyler1d: simpler!

src/actions/sessionActions.js
```diff
@@ -19,29 +17,7 @@ export function unRecognize() {
   }
 }
 
-export function loadUserSession() {
```
schuyler1d: folding in with token loading, so just deleting this one.

src/reducers/index.js
```diff
@@ -36,18 +36,23 @@ const initialSearchState = {
 
 const initialUserState = {
   // Possible keys. none of these are guaranteed to be present/available
-  // anonymous: <boolean>
+  // # OSDI fields:
```
schuyler1d: thought I'd try to organized the docs a little here.

src/reducers/index.js
```diff
@@ -227,9 +238,8 @@ function userReducer(state = initialUserState, action) {
           })
         }
       }
-    // No break here: fall through to the failures
+      return Object.assign({}, state, newData)
```
schuyler1d: without 4 cases, it's clearer and simpler just to keep all the case's distinct.  It was probably a bad idea to use fall-through in the first place.

PR #157, opened 2018-01-18
Using object-rest-spread, cleanup reducers a bit

gnarf:
Just some proposals to start trying to cleanup the reducers a bit.

object-rest-spread is such a benefit than Object.assign everywhere to readability IMO.

I also extracted the "nav" bits of the reducer into it's own file, in an attempt to show how the redux best practice suggest writing your reducers.  Happy to finish with the rest of them if you like how that looks compared to the old `navReducer`


schuyler1d:
Haven't reviewed yet to validate, but I love this and am :+1:  to adding.  More of the team should weigh  in on bringing in the new babel rule.

Thank you!

sreynen:
I also haven't tested, but 👍 on the general idea. Curious why the new file is named `reducers/nav/index.js` rather than `reducers/nav.js`.

gnarf:
> I also haven't tested, but 👍 on the general idea. Curious why the new file is named reducers/nav/index.js rather than reducers/nav.js.

I was gonna split up the properties even further, but then it was a small enough file that I've already made that change locally while writing the tests :)

PR #159, opened 2018-01-19
Nav selectors

gnarf:
This introduces a new concept in your redux implementation called "selectors"

Selectors in general allow you to write simpler `mapStateToProps` methods, they generally take the form of `(state, ...otherParams) => state.someKey`

This also allows you (as you can see in the second commit) to rename the location of storage in the redux state, and so long as you only ever use selectors to query the state, will ensure that we can change how the information is stored, and optimize the selectors.

In order to properly use this abstraction layer, we should never read properties directly from redux state, instead use `someSelector(state)` to get at the information we need, writing a selector if need be.

### DIFF
--- a/src/selectors/nav.js
+++ b/src/selectors/nav.js
@@ -1,5 +1,5 @@
 
-export const navOrg = (state, org) => state.nav.orgs && state.nav.orgs[org]
+export const navOrg = (state, org) => state.nav && state.nav.orgs && state.nav.orgs[org]
 export const cobrand = (state) => state.nav.partnerCobrand
 export const orgCobrand = (state, org) =>
   (org ? navOrg(state, org) : cobrand(state))
--- a/src/selectors/nav.js
+++ b/src/selectors/nav.js
@@ -1,0 +1,6 @@
+
+export const navOrg = (state, org) => state.navStore.orgs && state.navStore.orgs[org]
+export const cobrand = (state) => state.navStore.partnerCobrand
+export const orgCobrand = (state, org) =>
+  (org ? navOrg(state, org) : cobrand(state))
+
--- a/src/selectors/nav.js
+++ b/src/selectors/nav.js
@@ -1,6 +1,6 @@
 
-export const navOrg = (state, org) => state.navStore.orgs && state.navStore.orgs[org]
-export const cobrand = (state) => state.navStore.partnerCobrand
+export const navOrg = (state, org) => state.nav.orgs && state.nav.orgs[org]
+export const cobrand = (state) => state.nav.partnerCobrand
 export const orgCobrand = (state, org) =>
   (org ? navOrg(state, org) : cobrand(state))
 
--- a/src/actions/navActions.js
+++ b/src/actions/navActions.js
@@ -1,6 +1,8 @@
 import 'whatwg-fetch'
+import { navOrg } from '../selectors/nav'
 
 import Config from '../config.js'
+
 
 export const actionTypes = {
   FETCH_ORG_REQUEST: 'FETCH_ORG_REQUEST',
@@ -8,31 +10,30 @@ export const actionTypes = {
   FETCH_ORG_FAILURE: 'FETCH_ORG_FAILURE'
 }
 
-export function loadOrganization(orgSlug, forceReload) {
-  const urlKey = `organizations/${orgSlug}`
+export function loadOrganization(slug, forceReload) {
+  const urlKey = `organizations/${slug}`
   if (global && global.preloadObjects && global.preloadObjects[urlKey]) {
     return (dispatch) => {
       dispatch({
         type: actionTypes.FETCH_ORG_SUCCESS,
         org: window.preloadObjects[urlKey],
-        slug: orgSlug
+        slug
       })
     }
   }
   return (dispatch, getState) => {
     dispatch({
       type: actionTypes.FETCH_ORG_REQUEST,
-      slug: orgSlug
+      slug
     })
-    const { navStore } = getState()
-    if (!forceReload
-        && navStore
-        && navStore.orgs
-        && navStore.orgs[orgSlug]) {
+    const state = getState()
+    const org = navOrg(state, slug)
+
+    if (!forceReload && org) {
       return dispatch({
         type: actionTypes.FETCH_ORG_SUCCESS,
-        org: navStore.orgs[orgSlug],
-        slug: orgSlug
+        org,
+        slug
       })
     }
     return fetch(`${Config.API_URI}/api/v1/${urlKey}.json`)
@@ -41,14 +42,14 @@ export function loadOrganization(orgSlug, forceReload) {
           dispatch({
             type: actionTypes.FETCH_ORG_SUCCESS,
             org: json,
-            slug: json.name || orgSlug
+            slug: json.name || slug
           })
         }),
         (err) => {
           dispatch({
             type: actionTypes.FETCH_ORG_FAILURE,
             error: err,
-            slug: orgSlug
+            slug
           })
         }
       )
--- a/src/actions/navActions.js
+++ b/src/actions/navActions.js
@@ -3,7 +3,6 @@ import { navOrg } from '../selectors/nav'
 
 import Config from '../config.js'
 
-
 export const actionTypes = {
   FETCH_ORG_REQUEST: 'FETCH_ORG_REQUEST',
   FETCH_ORG_SUCCESS: 'FETCH_ORG_SUCCESS',
--- a/src/components/nav.js
+++ b/src/components/nav.js
@@ -1,12 +1,11 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
+import { orgCobrand } from '../selectors/nav'
 
 import NavLink from './nav-link'
 
-const Nav = ({ user, nav, organization }) => {
-  const cobrand = ((organization) ? nav.orgs[organization] : nav.partnerCobrand)
-
+const Nav = ({ user, cobrand }) => {
   const userLinks = (
     <div className='pull-right bump-top-1 span-7 top-menu'>
       <ul className='nav collapse nav-collapse'>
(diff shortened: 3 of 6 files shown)


lady3bean:
This pattern clears up what we need to do in each map state to props in components - here's what I see happening - 

- Instead of passing in the whole nav store in map state to props, we pass in state (which makes more sense of the function name), and we use selectors to grab the state of the only properties we care about. In this case, in the nav component we only need the state of cobrand, and on the homepage we only need navorg. This way we import the selectors we need for each component, rather than updating the component for every state change for all properties in the store.

- I like how this organizes the logic of the selectors out from components. I think this will be helpful in keeping track of what properties we are getting from the store for each reducer. And then it's also clearer when we use these props in a component.

Definitely keen on seeing how this way of organizing the code would play out for the user and petition reducers

gnarf:
Happy to carry the pattern through myself for one big cleanup commit, it might cause some merge conflicts if other people are working on redux'ed components, but I'm happy to help solve those too!

schuyler1d:
I guess I liked them being named `navStore` or `*Store` since we import so many TYPES of things, that being clear that it's the redux store improved readability for me -- especially with a team that is not constantly working in React (we might be now, but less later when we come back to it), the different `state`, `props`, `store`, etc get confusing and being able to follow the wires through the variable names is important to me.

I need to look at exactly what the selector pattern is/is-doing

schuyler1d:
ok, I think I get selectors now.  I suppose the benefit is separating the selector from the component, but it's not clear to me why we (for this project, anyway) need that additional layer of indirection.  the `mapStateToProps` pattern/function already contains that code in a specific and functional/pipeline way.  If we had selectors that were used across multiple components, then I see where that's valuable.

I think I'm not enthusiastic about this direction for this project at this time -- though I see the usefulness for others.  I will let others weight in -- I'm happy to be overridden/out-voted.

gnarf:
> I guess I liked them being named navStore or *Store since we import so many TYPES of things,

except you're getting a `state` not a `store` -- store is an object that has `.getState()` and `.dispatch()` methods -- `state.nav` is pretty unambiguous IMO

> that being clear that it's the redux store improved readability for me

redux state only being read by selectors makes the pretty clear -- you get the properties you want, and only the properties you want by using selectors, those become `props` in the react component.  If you choose to use the components `state` as well, that can become confusing, but since you only ever deal with the redux state in `connect(mapStateToProps)`, the "location" of the information is straightforward

schuyler1d:
Yes, I know it's a redux state, but that's incredibly confusing (again to occasional/light React developers) against the `state` in a component.  I purposefully renamed the redux 'state' to store, to make that separation clearer, and I stand by that.  (Again, others are welcome to weigh in).

I see that maybe confusing the redux store to the redux state could also be confusing.  But whereas people are reading mapStateToProps in lots of places, we're rarely interacting with the store directly -- we are simply `connect`ing and getting the `dispatch` function and everything else is hidden.

gnarf:
> it's not clear to me why we (for this project, anyway) need that additional layer of indirection

The reasoning is a few fold...

1. If you only use selectors to access redux state = less confusion about which "state" we are talking about
2. Performance - Currently you have some components doing `{ user: state.userStore }` as a property in the mapStateToProps, this can become very taxing as every time any action mutates anything in the `userStore` every component using that property will be re-rendered, even if it didn't change the data.  You'll save long term on performance by mapping only what you need like `(state)
 => ({ anonymous: isAnonymous(state),  givenName: givenName(state) })`
3. It's a very common practice -- there are multiple libraries that will work with these sorts of selectors, like reselect which allows for caching "generated" properties.

gnarf:
Also, if you ever change anything about how the user sessions are stored, you'll only need to update your selectors file, not every component accessing the redux state.

gnarf:
> but that's incredibly confusing (again to occasional/light React developers) against the state in a component

I agree that it can be confusing at first, but since you only access redux state inside `connect(mapStateToProps)` I feel that it's pretty straight-forward that `state` in that context is the redux state.  `component.state` is a completely different beast.

We shouldn't ever be passing `state` itself into the component, just properties selected from the state, so you shouldn't ever end up with the extra-confusing `this.props.state` ;)

gnarf:
> If we had selectors that were used across multiple components

You already have this problem.

Every one of these components is accessing `state.userStore`, this is a spreading problem that will make it impossible to ever refactor/clean up your redux state without the selectors abstraction

* components/nav
* actions/petitionActions
* components/signature-add-form
* pages/search.js
* pages/thanks.js

5 places that all access the same data

schuyler1d:
I don't think I get the performance argument.  it's the same obj key navigation in the function or directly.

The state isn't updated when it's not changed, e.g. for userStore here (and below):
https://github.com/MoveOnOrg/mop-frontend/blob/master/src/reducers/index.js#L229

Perhaps the objection/issue is that we're passing the 'whole' user state rather than pieces of it.  In general, I agree with that -- and you see that we don't do that for the others -- i.e. just send the relevant `petition`, etc.  I think sending minimal state to the components is a good practice -- userStore is small enough and all-important, that any/all changes to it *should* update all places it's needed/used.

I suppose we could protect that with a sub-key for userStore, but -- that's also why it's nice to have the variable called `userStore` -- very easily greppable.

gnarf:
perhaps a "pro tip" from the author of redux will help convince this is a good idea in the long run.

https://twitter.com/dan_abramov/status/664581975764766721?lang=en

> They let you decouple views and action creators from state shape

schuyler1d:
Yeah, I get the decoupling argument.  We're just not there yet -- I don't think we have a single component that's used in more than one place (ok, state-selection, but that's basically stateless and way-down).  When we do (maybe when our Reviewer work or a Preview mode comes in), then I'll be in favor of decoupling, and I'm glad there's a standard pattern called selectors when we get there.

schuyler1d:
btw, if you are interested in helping on all-things-React, we have another project:
https://github.com/MoveOnOrg/Spoke
which needs a LOT more love.  We don't even have a working test framework for the frontend (way back, when I had a chance to try something, I think it was blocked on `enzyme` not supporting React 15.6).  It couuld benefit from a React 16 upgrade, ... 
:heart:

gnarf:
> We're just not there yet

You already have 5 files tied to the exact state shape to get at user data (and I imagine this will grow quickly). 

What happens when you decide later that the user data storage makes more sense as a part of a broader `session` reducer, you'll be stuck hunting through every file (for your luckily uniquely named `userStore` 😸 ) property and converting it then :)

gnarf:
> if you are interested in helping on all-things-React

Happy to help wherever, I can be a refactor-bot.  Started here cuz it's the project @lady3bean keeps talking to me about ;)

lady3bean:
I find the selectors help me reason about what in the react framework is responsible for what.

Though I see what you're saying Sky about each component having a pipeline, and that it therefore makes sense to do the logic of what the component needs from the redux store in each component, it also feels like it adds complexity to components. Separating out the selectors and importing the ones you need into components make the components as well as how we're using the store to update the properties we care about both more readable. If we really wouldn't re-use the same selector across different components I can see how this pattern might be over-optimizing things, but it just makes more sense to me separated out - selectors kind of remind me of having more of an mvc structure, rather than components being responsible for defining its redux pipeline. With this PR, I can see what we need from the store for nav generally, and then in each component, it's more clear what props are being used.

gnarf:
just rebased on master

sreynen:
After reading through this, I generally agree with Sky that this doesn't seem valuable as a universal pattern. To address the specific points:

1. I don't see how selectors reduce confusion on component vs. redux state. The confusion there stems from the ambiguous parameter coming into `mapStateToProps`, and selectors don't change that, only abstract how we delve into the shape of that parameter.

2. I also don't see how selectors improve performance. Re-rendering is triggered by prop updates, so the performance improvement comes from mapping props more specifically, which can be easily done without selectors. And that actually seems easier to do without selectors, because it can be done more specific to each individual component's prop needs.

3. A practice being common doesn't really tell us anything about its value.

4. Being able to update a complex and/or frequently-used state shape in one place seems like the main benefit of selectors, or more generally a benefit to abstracting functionality in functions, which seems to be all selectors are doing. That's how it's used in the examples Dan Abramov points to, and those Redux examples don't use selectors everywhere, so it seems like the best practice with selectors is to use selectors only sometimes, not universally.

All of that is my argument against using selectors everywhere, but that's not even a thing we need consensus on for this individual PR, just some context for future PRs. For this individual PR, we only need to consider the 3 selectors added: `navOrg`, `cobrand`, and `orgCobrand`. `navOrg` is used in two places and removes some redundant code, so that seems useful.  `cobrand` is only used in one place, that place is another selector, plus it's only returning a single value, so that doesn't seem like useful layer of abstraction. `orgCobrand` is only used in one place, but it's complex enough that breaking it out to its own function seems worthwhile.

So my suggestion is to remove `cobrand`, merge the rest, and not treat selectors as a universal pattern, just a thing to consider in specific cases.

src/selectors/nav.js
```diff
@@ -0,0 +1,6 @@
+
+export const navOrg = (state, org) => state.nav && state.nav.orgs && state.nav.orgs[org]
+export const cobrand = (state) => state.nav.partnerCobrand
```
sreynen: This function doesn't seem useful currently, so suggest removing.

schuyler1d:
So after some discussion, we're going to close this PR.

As a team, we are pro-reducers but only on a case-by-case basis when:
* the logic is used in more than one component
* the logic is complex -- goes beyond traversing into an object (which can just use ES6 destructuring)
* the logic isn't better done by refactoring the store data in the first place

These three reducers don't meet that criteria, but I'm really happy that we are conscious of the pattern now as a team.  Thank you!

gnarf:
*shrug* 2+ years of working with react (since 0.13) and redux (since version 1) tells me the decision not to start down this route now, while things are still simple, is gonna come around to bite you in the butt.  Just because my first example of doing this with the "nav" reducer/state didn't get much "reuse" doesn't mean that the pattern isn't going to simplify the code in the long run, or help you reason about whats happening.

I understand my arguments for the practice aren't particularly compelling other than "I've done it in many react projects, and it helped everyone I've ever worked with on a react project reason about the state flow" - it's because it really is such a "small" thing, that it doesn't make much of a difference, until it does, and then it becomes a huge refactoring trap.

Besides, there are a lot of other places that can be cleaned up first, and this isn't really that hard of a refactor to do again some day later.

gnarf:
I've added a few commits to this branch, to try to show how the selectors can be more useful than my first PR'ings here showed, but they don't seem to append to a closed PR.

The general changes I made:

* Added some documentation to the selectors, showing the shape and describing what they should retrieve
* Rewrote the reducer tests to use the selectors (was planning to do this in a followup anyway, but hadn't yet)

schuyler1d:
It seems when an issue is closed new commits don't come in, but to review it, I'm dropping a link to see the new diff:
https://github.com/MoveOnOrg/mop-frontend/compare/master...gnarf:nav-selectors?expand=1

PR #161, opened 2018-01-19
Switch tests for Home to be shallow and not use redux

gnarf:
**This PR requires #159 to land first, but wanted to get this submitted sooner rather than later.**

**Pay the most attention to the last commit 7a0afa7 during review.**

This switches the tests for `<Home />` to shallow test only the inner component. the connect / redux bits don't need to be tested as deeply, but if you wanted, you could also just test the `mapStateToProps` function.

By exporting the `WrappedComponent` without the connect wrapper, we can completely avoid mounting deeply, or with redux states at all and worry about what we want the component to present when given data.

After this change, these 4 tests that I swapped over are already saving ~75ms each just switching to shallow instead of mount, and avoiding setting up stores.

shallow / reduxless tests
```
  4 passing (165ms)
  4 passing (160ms)
  4 passing (142ms)
  4 passing (144ms)
  4 passing (139ms)
  4 passing (146ms)
  4 passing (140ms)
  4 passing (141ms)
  4 passing (146ms)
  4 passing (140ms)
```

vs mount/provider/store tests

```
  4 passing (508ms)
  4 passing (437ms)
  4 passing (352ms)
  4 passing (502ms)
  4 passing (354ms)
  4 passing (356ms)
  4 passing (351ms)
  4 passing (361ms)
  4 passing (363ms)
  4 passing (343ms)
```

If y'all wanna test the times on your own machines, here's the command I used to generate the output:
```bash
for i in $(seq 1 10); do mocha --compilers js:babel-core/register --require jsdom-global/register --require isomorphic-fetch test/pages/home.js  | grep passing ; done
```

### DIFF
--- a/src/actions/navActions.js
+++ b/src/actions/navActions.js
@@ -3,6 +3,7 @@ import { navOrg } from '../selectors/nav'
 
 import Config from '../config.js'
 
+
 export const actionTypes = {
   FETCH_ORG_REQUEST: 'FETCH_ORG_REQUEST',
   FETCH_ORG_SUCCESS: 'FETCH_ORG_SUCCESS',
--- a/src/actions/navActions.js
+++ b/src/actions/navActions.js
@@ -1,6 +1,8 @@
 import 'whatwg-fetch'
+import { navOrg } from '../selectors/nav'
 
 import Config from '../config.js'
+
 
 export const actionTypes = {
   FETCH_ORG_REQUEST: 'FETCH_ORG_REQUEST',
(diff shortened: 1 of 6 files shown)


schuyler1d:
Can you rebase this off of master, so the reducer PR isn't a dependency?  Once that's true, I'm :+1:

gnarf:
> Can you rebase this off of master, so the reducer PR isn't a dependency?

It would be a bit of work to do that, I used a bit of the selectors refactoring in the tests.  If we are going to land selectors, I'd prefer to not have to rebase again, but if we are going to just close the selectors refactor, I'll just rewrite this one later

gnarf:
gotta rewrite this without my other refactors if you still want it

PR #162, opened 2018-01-19
md5Message was not propagating to thanks page. 

schuyler1d:
This fixes, and also makes it possible to test locally with API_WRITABLE=fake


schuyler1d:
We could though we haven't done any testing of actions yet.  If we were going to add tests, I would probably suggest making it a function argument as well, so the test shouldn't have to change environment variables.

lady3bean:
Ok. I just didn't know if there was anything we should do for approving this PR - it makes sense though.

PR #166, opened 2018-01-19
Thanks page finishings

schuyler1d:
This implements two things that were marked TODO:
* passing the  ?from_source= parameter from the sign page to the thanks page and having that change the `pre` variable
* doing some length logic around really long petition statements for the mail message.

This is on top of PR: https://github.com/MoveOnOrg/mop-frontend/pull/165
so we should merge that first, and then I'll rebase this.

### DIFF
--- a/src/components/thanks.js
+++ b/src/components/thanks.js
@@ -82,16 +82,25 @@ class Thanks extends React.Component {
   }
 
   render() {
-    const { petition, user, signatureMessage } = this.props
-    const creator = false // Maybe test user.id==petition.creator_id or something, if we want to expose that
-    const pre = (creator ? 'c' : 's') // TODO: based on ?from_source= parameter .icn and .imn, megapartner
-    const actedOn = (creator ? 'created' : 'signed')
+    const { petition, user, signatureMessage, fromSource } = this.props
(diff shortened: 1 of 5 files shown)


src/components/thanks.js
```diff
@@ -106,31 +134,23 @@ class Thanks extends React.Component {
...
-
-Thanks!
-`)
-    const copyPasteMessage = `Subject: ${petition.summary}\n\n${mailToMessage}` // TODO: replace __TYPE__
+    const mailToMessage = this.generateMailMessage(textDescription.textContent,
+                                                   petition.summary,
+                                                   isCreator,
+                                                   shareOpts,
+                                                   petition.target,
+                                                   `${link}?source=${pre}.em.__TYPE__&${this.trackingParams}`
```
sreynen: It looks like both `link` and `pre` are only used to compose this param to `generateMailMessage`. Unless those will eventually be used elsewhere, suggest replacing the param with whatever is needed to create `link` and `pre` (which is maybe just the whole petition object?) and move that logic to `generateMailMessage`.

schuyler1d: Thanks --that actually exposed an issue -- i was assuming `this.state.pre` in a bunch of places, but it hadn't been set.

PR #169, opened 2018-01-23
Flag comment

schuyler1d:
Flagging a comment notifies the backend.  This loads in a comment_id field from the api, and then makes an interface to send a flag to the backend.  This it gives feedback that it was done.

This has a couple small drive-by touches on search stuff to make tests pass without warnings.


src/actions/petitionActions.js
```diff
@@ -22,7 +22,10 @@ export const actionTypes = {
 
   PETITION_SIGNATURE_SUBMIT: 'PETITION_SIGNATURE_SUBMIT',
   PETITION_SIGNATURE_SUCCESS: 'PETITION_SIGNATURE_SUCCESS',
-  PETITION_SIGNATURE_FAILURE: 'PETITION_SIGNATURE_FAILURE'
+  PETITION_SIGNATURE_FAILURE: 'PETITION_SIGNATURE_FAILURE',
+
+  FEEDBACK_SUCCESS: 'FEEDBACK_SUCCESS',
```
schuyler1d: Since we don't really care too much whether the feedback made it, we don't process these at all or distinguish besides additional action parameters feedback for the comment vs. petition.  But if we do care at some point, they'll be here.

src/actions/petitionActions.js
```diff
@@ -303,6 +306,60 @@ export const loadPetitionSignatures = (petitionSlug, page = 1) => {
   }
 }
 
+export const flagPetition = (petitionId, reason) => (dispatch) => {
```
schuyler1d: these are both close to identical but with slightly different urls and arguments.

src/components/petition-flag-form.js
```diff
@@ -70,7 +71,9 @@ class PetitionFlagForm extends React.Component {
...
 
 PetitionFlagForm.propTypes = {
-  petition: PropTypes.object.isRequired
+  petition: PropTypes.object.isRequired,
+  dispatch: PropTypes.func
 }
 
-export default PetitionFlagForm
+export default connect()(PetitionFlagForm)
+export const WrappedComponent = PetitionFlagForm
```
schuyler1d: started using gnarf's pattern here, since it makes testing much easier.

src/components/search-results.js
```diff
@@ -15,7 +15,7 @@ class SearchResults extends React.Component {
 
   render() {
     const { searchResults, currentPage, query } = this.props
-    const resultCount = searchResults.count
+    const resultCount = Number(searchResults.count || 0)
```
schuyler1d: this was a bug showing up in the tests.  The api (oddly?) sends the count as a string.  I won't argue with that, but then we need to convert it to a number here.

src/components/signature-list-item.js
```diff
@@ -1,30 +1,62 @@
...
-import { formatDate } from '../lib.js'
+import { formatDate } from '../lib'
+import { flagComment } from '../actions/petitionActions'
 
-const SignatureListItem = ({ user, createdDate, number, comments }) => {
-  const { city, state, name } = user
-  const location = ((!city) ? state : `${city}, ${state}`)
-  const fromLocation = ((location === '') ? '' : `from ${location}`)
-  const date = new Date(createdDate)
+class SignatureListItem extends React.Component {
```
schuyler1d: This got state, so it needed to be become a full component.

src/components/signature-list-item.js
```diff
@@ -1,30 +1,62 @@
...
+          <b>{name}</b> {fromLocation} signed this petition on {formatDate(date)}.
         </div>
-        : null
-      }
-    </li>
-  )
+        {comments ?
+          <div className='quoteboxup'>
+            {this.state.flagged
+             ? // clicked to flag
```
schuyler1d: The old system makes this a whole page rather than just updating with a note, but I thought the immediate feedback and experience was better anyway, so hopefully we don't see this as a regression.

src/components/signature-list-item.js
```diff
@@ -33,9 +65,12 @@ SignatureListItem.propTypes = {
     state: PropTypes.string.isRequired,
     name: PropTypes.string.isRequired
   }),
-  createdDate: PropTypes.number.isRequired,
+  createdDate: PropTypes.string.isRequired,
```
schuyler1d: the api sends a string.

src/pages/search.js
```diff
@@ -18,7 +18,6 @@ const SearchPage = ({ query, selectState, pageNumber }) =>
   </div>
 
 SearchPage.propTypes = {
-  user: PropTypes.object,
```
schuyler1d: `user` isn't used at all in search -- it's just a copy/paste typo

test/components/signature-list-item.js
```diff
@@ -3,15 +3,15 @@ import { expect } from 'chai'
 
 import { shallow } from 'enzyme'
 
-import SignatureListItem from '../../src/components/signature-list-item'
+import { WrappedComponent as SignatureListItem } from '../../src/components/signature-list-item'
```
schuyler1d: See!  I didn't have to rewrite the tests.  This is great.

mathemagica: 👍

PR #178, opened 2018-01-24
113 signature add form tests

lady3bean:


lady3bean:
So, after some misadventures with trying to use createMockDispatch in order to simulate clicks on logging out and submitting the form, I went with using enzyme's actual click simulate function. This does seem to call the function I'm expecting - and maybe it's worth doing this for logging out, which so far I've treated by testing what renders based on state, rather than clicking the function.

Overall though, testing the functionality of petition and session actions seems like it belongs in separate tests, not here. I think we should be testing what the component renders based on state, not how redux is handling state changes.

Feedback on the strategy here certainly welcome though! I want to make sure I'm testing the appropriate things there.

Issue #175, opened 2018-01-23
Search bar on search page has extra elements


sreynen:
Should have no state select, nor "Search for petitions by any keyword" prompt. See https://petitions.moveon.org/find/

shakalee14:
i configured the search bar so it can be long or short, so this should be an easy fix

PR #179, opened 2018-01-24
Adds optional render for short or long search bar

shakalee14:


shakalee14:
for #175

schuyler1d:
please run `npm run lint -- --fix` and fix the lint errors.

I seem to remember something where it turned out there was never a short search bar and that's why it wasn't implemented?  @mathemagica ?

I'd recommend if we DO do this, then there should be just an `isLong` (or isShort) boolean property, so you don't have to match on a string.

shakalee14:
this should have had wip on it. but i'm trying to fix for the fact that the search bar looks different on the search page vs the home page. I call one short because it has less attributes (no state select). I don't have to call it short.

shakalee14:
Note, this is adding search.css (Which is on the legacy site). So that needs to be deployed too.

mathemagica:
@schuyler1d Shaka's original implementation had the long vs short bars, and then I merged them into one because I missed that http://petitions.moveon.org/find/ didn't have a state selector and didn't think the separate cases were needed- my bad.

So I think this implementation is right, and ++ to making the property boolean whatever its called. And also could you add a test for long vs short in tests/components/search-results.js ?

shakalee14:
@mathemagica 

made it boolean with this commit: https://github.com/MoveOnOrg/mop-frontend/pull/179/commits/8b6552af5f6e356b77cc1591d157213023c36cf3

added test with this commit: 

https://github.com/MoveOnOrg/mop-frontend/pull/179/commits/81aec5ce7ce1bc1da87debb7b07920b8b659800e

css/moui.css
```diff
@@ -8566,3 +8566,44 @@ body.donate label.label-left {
     width: 97%;
   }
 }
+
+/*previously search css - css selectors are only found here*/
+
+.result {
```
schuyler1d: Can you add an additional class here:
https://github.com/MoveOnOrg/mop-frontend/blob/master/src/pages/search.js#L9
like `search-page` or whatever, and then prefix all of these with `.search-page` -- e.g. `.search-page .result { ...`

Issue #187, opened 2018-01-25
General layout on thanks page is wrong


sreynen:
Columns are too far apart, and email textarea is too short.

PR #190, opened 2018-01-25
187 thanks page formatting

shakalee14:


shakalee14:
for #187

shakalee14:
I also tried to make email language look more like what's on prod, using the following example:
http://petitions.moveon.org/sign/keep-hearst-elementary
vs.
localhost:8080/#/sign/keep-hearst-elementary

src/components/thanks.js
```diff
@@ -43,11 +42,6 @@ class Thanks extends React.Component {
     petitionActions.recordShareClick(this.props.petition, medium, source, this.props.user)
   }
 
-  shareLink(evt) {
```
schuyler1d: I know this isn't part of prod, but can we keep this -- it drives me crazy that I can't get a petition link from the thanks page.

src/components/thanks.js
```diff
@@ -137,14 +130,14 @@ Thanks!
...
-                                                   petition.summary,
+                                                   petition.title,
                                                    this.isCreator,
                                                    shareOpts,
                                                    petition.target,
                                                    `${petition._links.url}?source=${this.state.pre}.em.__TYPE__&${this.trackingParams}`
                                                    )
-    const copyPasteMessage = `Subject: ${petition.summary}\n\n${mailToMessage.replace('__TYPE__', 'cp')}`
-    const mailtoMessage = `mailto:?subject=${encodeURIComponent(petition.summary)}&body=${encodeURIComponent(mailToMessage.replace('__TYPE__', 'mt'))}`
+    const copyPasteMessage = `Subject: ${petition.title}\n\n${mailToMessage.replace('__TYPE__', 'cp')}`
```
schuyler1d: this is definitely correct.

src/components/thanks.js
```diff
@@ -137,14 +130,14 @@ Thanks!
     }
 
     const mailToMessage = this.generateMailMessage(textDescription.textContent,
-                                                   petition.summary,
+                                                   petition.title,
```
schuyler1d: You're right that we need the title as well as the summary, but it shouldn't universally replace it.  You are covering this case:
https://github.com/MoveOnOrg/motrunk1/blob/soak/www/pac/signon/share.mc#L155

but see the other cases above that

PR #192, opened 2018-01-25
scroll to top on page change

### LINKED ISSUE #168: Give clearer visual indication search page has changed
This is kind of a problem created by things working _too_ well. When clicking through search pages, the results load fast and there's no page refresh, which is great. But as a result, it's not always clear the new page has loaded.

There are probably a lot of ways we could fix this, and this may not be the best way, but my initial thought is it should scroll back to the top when a new page loads.

sreynen:
Fixes #168. I don't feel like this is a great fix, but it seems good enough for now, a clear indication that something has changed while paging through results. Jumping back to the top is basically the same behavior we had before, but without reloading the whole page.


sreynen:
https://reactjs.org/blog/2017/09/26/react-v16.0.html says 
> Shallow renderer no longer calls componentDidUpdate because DOM refs are not available. 
I think it was deprecated in a specific context, but that doesn't seem to be the context in use here, because it still fires as desired here.

I played around with a few different scroll positions and top of the results div seems best.

sreynen:
@schuyler1d approved this in chat, so merging.

Issue #150, opened 2018-01-16
Create register page and component


sreynen:
Page should look like https://petitions.moveon.org/login/register.html and submit to API, but can be mock API for testing, since response is entirely server-side. After submit, should redirect to https://petitions.moveon.org/no_petition.html

PR #197, opened 2018-01-26
Add Register Page and Component WIP

caleblogan:
#150

Not really sure about the server api so I just loosely filled in the actions and reducers.  Would it be better to incorporate the register reducer with the session reducer?

### DIFF
--- a/src/actions/accountActions.js
+++ b/src/actions/accountActions.js
@@ -1,0 +1,48 @@
+import 'whatwg-fetch'
+
+import Config from '../config.js'
+
+export const actionTypes = {
+  REGISTER_SUBMIT: 'REGISTER_SUBMIT',
+  REGISTER_FAILURE: 'REGISTER_FAILURE',
+  REGISTER_SUCCESS: 'REGISTER_SUCCESS'
+}
+
+export function register(fields) {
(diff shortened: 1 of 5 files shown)


shakalee14:
This looks great! While looking through this branch (in comparison to legacy system), I noticed three things. I think the first one should be handled by the client.  

1) In the legacy system, there is a check for a valid email in the form. I think we need to add this logic for this PR. So, see the following:

screen shot 2018-01-29 at 10 34 28 am

2) It looks like email and name are updated in the url for an invalid email, notice (this may not be handled in the client, but adding that note):
https://petitions.moveon.org/login/register.html?email=dkafsdj&name=Love

3) If email is already registered, form error changes to 'A account already exists for your email address on the MoveOn petition website.'. I think this will be handled server-side.

caleblogan:
Thanks for reviewing!

1) I added the regex check found in SignatureAddForm for validating emails. It may be a good idea to either use a form library or create a module for all the form validators.

2) I believe that is from an http redirect to repopulate the form if there are errors. Since state is held in the client now that shouldn't be needed.

3) Yeah, I think that error should come from the server.

shakalee14:
LGTM!

Issue #186, opened 2018-01-25
Flagging a signature comment needs confirmation dialog


sreynen:
Current site has a modal "Flag this comment as inappropriate?" to confirm the action. Need that on the new version.

jparkrr:
Seems like this can be closed @sreynen ?

sreynen:
Yep.

PR #198, opened 2018-01-27
Add Confirmation Dialog to Flag Signature Comment

caleblogan:
#186 

I did what the live site does and used the browser confirm() method.  Would it be better to create a modal component and do something like https://getbootstrap.com/docs/3.3/javascript/#modals?


caleblogan:
Not yet. I will check it out though. I found your project on the reactiflux discord channel.

PR #199, opened 2018-01-29
create petition page

### LINKED ISSUE #55: Add petition creation form component, add to petition create page
For deployment this will be blocked on: https://github.com/MoveOnOrg/mop/issues/23
however, part of this should be to create a set of file(s) in `./local/api/v1/targets/{group}.json` which can be used for testing.  You can assume the output of target_search.html but you can also require additional (useful?) fields that the backend can use.

sreynen:
Fixes #55. This doesn't entirely complete #55, but we're going to create separate issues for what's missing.


src/components/create-petition-form.js
```diff
@@ -0,0 +1,226 @@
...
+      selected: 'title',
+      nationalOpen: false,
+      stateOpen: false,
+      customOpen: false
+    }
+  }
+
+  render() {
+    const { selected, nationalOpen, stateOpen, customOpen } = this.state
+    const elementByField = {
```
sreynen: This maps element identifiers to the element reference created with `ref=` attributes below.

src/components/create-petition-form.js
```diff
@@ -0,0 +1,226 @@
...
+      title: this.titleInput,
+      statement: this.statementInput,
+      'target-national': this.nationalInput,
+      'target-state': this.stateInput,
+      'target-custom': this.customInput,
+      about: this.aboutInput
+    }
+    const instructionsByField = {
```
sreynen: This is a list of instructions for each element.

src/components/create-petition-form.js
```diff
@@ -0,0 +1,226 @@
...
+            <input name='text_custom_group_title' id='custom_title' className='text' type='text' placeholder='Title or Position (optional)' />
+          </div>
+          <div id='someone_else_add'>
+            <a href='#' className='btn' id='add_this_target'>Add another target</a>
+          </div>
+        </div>
+      </div>
+    )
+    const arrowStyle = { position: 'relative', top: 88 }
+    const instructionStyle = { position: 'relative', top: -45 }
```
sreynen: Instructions and arrow move with selected element, so using inline styles for that part.

src/components/create-petition-form.js
```diff
@@ -0,0 +1,226 @@
...
+            <form id='petition_form'>
+              <input value='' name='targets' id='targets_json' type='hidden' />
+              <input value='' name='skin' type='hidden' />
+              <input value='' name='source' type='hidden' />
+              <input value='' name='cloned_from_id' type='hidden' />
+              <input value='' name='solicit_id' type='hidden' />
+              <fieldset id='start'>
+                <span className='circle-number'>1</span><span className='lanky-header moveon-dark-blue'> Start your petition!</span>
+                <div className='text wrapper big' id='text_name_wrapper'>
+                  <input name='name' id='name_field' className='span6' type='text' title='Your Petition Title' placeholder='Petition title' onClick={() => this.setState({ selected: 'title' })} ref={(input) => { this.titleInput = input }} />
```
sreynen: `onClick` sets the selected state, and `ref` allows React to grab the location of the selected element, to update the content and position of instructions.

src/components/form/state-select.js
```diff
@@ -1,12 +1,15 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 
-const StateSelect = ({ onChange, selectText, style }) => (
+const StateSelect = ({
+  onChange, selectText, style,
+  name = 'state', id = 'state_id', className = 'span4 state moveon-track-click'
```
sreynen: Reuse state select from sign form, but with different attributes.

src/components/create-petition-form.js
```diff
@@ -0,0 +1,226 @@
...
+          <h4>Text of your Petition</h4>
+          <p>This is the message that petition signers are actually signing, to be delivered to the decision-maker. You will get a lot more signers if your message is short and sweet&mdash;one or two sentences at the most. The petition statement is NOT the place to make a detailed persuasive argument for your position.</p>
+          <p><strong>Example:</strong> &ldquo;Stop the proposed cuts in funding for Dewey Elementary School.&rdquo;</p>
+        </div>
+      ),
+      'target-national': (
+        <div>
+          <h4>Targeting the White House or Congress</h4>
+          <p>If you choose <strong>The entire U.S. House</strong>, then your petition signers will be asked to sign a petition addressed to their individual representative in the U.S. House of Representatives.</p>
+          <p>If you choose <strong>The entire U.S. Senate</strong>, then your petition signers will be asked to sign a petition addressed to their state's U.S. Senators.</p><p>If you choose <strong>President Donald Trump</strong>, your petition signatures will be addressed to President Donald Trump.</p>
```
schuyler1d: Can we make this an `lsquo` just so github (and my editor :-) don't mess up the highlighting?

src/components/create-petition-form.js
```diff
@@ -0,0 +1,226 @@
...
+        <div>
+          <h4>Petition Background</h4>
+          <p>In 2-3 sentences, why should people be concerned about this issue? Is there a deadline? Have you been <i>personally</i> affected by the issue?</p>
+          <h4>Example:</h4>
+          <p>&ldquo;My daughter Maria attends Dewey Elementary School, which is about to lose 10 teachers and funding for activities like chess club. Our children's education should be our top priority, and these cuts should be stopped.&rdquo;</p>
+        </div>
+      )
+    }
+    const instructions = instructionsByField[selected]
+    const national = (!nationalOpen) ? '' : (
```
schuyler1d: I think between separate component methods (or components) and large consts, I'd prefer one of the former here.  I do like it separated out -- it'll make it more readable below in this/the render method.

src/components/create-petition-form.js
schuyler1d: if these are all true consts (not changed by context), can we put them somewhere other than inline of the render method?

sreynen:
I'm going to merge this so it's easier to work from on follow up tasks, but I do plan to come back and apply all the feedback.

PR #200, opened 2018-01-30
allows us to deploy parts of the app with prodReady <Route /> property,...

### LINKED ISSUE #63: Thanks page should hide some header links
A reduced header should be shown to focus people on thanks page actions.

This should probably be implemented in a way that ensures that the full header returns easily in other contexts.

### LINKED ISSUE #201: Hide most navigation on create petition page

- have the nav display handled with redux - we usually hide the nav when we don't want it to distract from the action featured on the page

schuyler1d:
...,but make ALL petitions.moveon.org links use <Link to= /> and then direct appropriately.

This should make it straightforward for us to distinguish what's 'done and ready' and still being worked on and allow us to deploy production pieces that have passed while some still-not-complete (or QA/tested) components, can still live in master with their routes.

This now also has a second commit -- see https://github.com/MoveOnOrg/mop-frontend/pull/200/commits/8825c11060837662462d1d1a8e516d82ae5e7047 for the second one, which fixes #201 and fixes #63

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,5 +8,8 @@
  * assuming redux-thunk, no mapPropsToDispatch -- just call `dipatch(actions.foo(data))`
  * `actions/*` is where you call out to get data (async or otherwise)
  * lazy-loading info: (TODO)
- 
-
+ * Even though convention calls redux 'state' 'state' -- we call it `store`, to map better to the redux
+   global store object, and distinguish it from component states.  c.f. any component that has a mapStateToProps function
+ * Any links to petitions.moveon.org should use react-router's `<Link to= />` regardless of it being in or out of the react app.
+   - The url must be absolute and not relative (start with a `/`)
+   - See src/routes.js for details -- when a route's app is 'production ready', we need to mark it with `prodReady` in the appropriate `<Route />`
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,5 +11,5 @@
  * Even though convention calls redux 'state' 'state' -- we call it `store`, to map better to the redux
    global store object, and distinguish it from component states.  c.f. any component that has a mapStateToProps function
  * Any links to petitions.moveon.org should use react-router's `<Link to= />` regardless of it being in or out of the react app.
-   - The url must be absolute and not relative (start with a `/`)
+   - The `to` must be an absolute path and not relative (i.e. start with a `/`)
    - See src/routes.js for details -- when a route's app is 'production ready', we need to mark it with `prodReady` in the appropriate `<Route />`
--- a/src/actions/petitionActions.js
+++ b/src/actions/petitionActions.js
@@ -169,7 +169,7 @@ export const registerSignatureAndThanks = (petition, signature) => () => {
     window.gtag('event', 'conversion', { send_to: petition.gtag_keys })
   }
   // 2. show thanks page
-  let thanksUrl = 'thanks.html'
+  let thanksUrl = '/thanks.html'
   if (petition.thanks_url) {
     if (/^https?:\/\//.test(petition.thanks_url)) {
       document.location = petition.thanks_url // FUTURE: maybe add some query params
--- a/src/components/footer.js
+++ b/src/components/footer.js
@@ -1,4 +1,5 @@
 import React from 'react'
+import { Link } from 'react-router'
 
 const Footer = () => (
   <div className='container footer'>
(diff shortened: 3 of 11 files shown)


CONTRIBUTING.md
```diff
@@ -8,5 +8,8 @@
  * assuming redux-thunk, no mapPropsToDispatch -- just call `dipatch(actions.foo(data))`
  * `actions/*` is where you call out to get data (async or otherwise)
  * lazy-loading info: (TODO)
- 
-
+ * Even though convention calls redux 'state' 'state' -- we call it `store`, to map better to the redux
+   global store object, and distinguish it from component states.  c.f. any component that has a mapStateToProps function
+ * Any links to petitions.moveon.org should use react-router's `<Link to= />` regardless of it being in or out of the react app.
```
schuyler1d: This is documenting what I'm setting up here.  the main action is all in routes.js

src/actions/petitionActions.js
```diff
@@ -169,7 +169,7 @@ export const registerSignatureAndThanks = (petition, signature) => () => {
     window.gtag('event', 'conversion', { send_to: petition.gtag_keys })
   }
   // 2. show thanks page
-  let thanksUrl = 'thanks.html'
+  let thanksUrl = '/thanks.html'
```
schuyler1d: This is one of many examples where we clean up all places where our code does not have absolute urls

src/components/footer.js
```diff
@@ -11,19 +12,19 @@ const Footer = () => (
         </div>
         <div className='row'>
           <div className='span4'>
-            <a className='icon-start icon-link-narrow' href='https://petitions.moveon.org/create_start.html?source=bnav'>Start a Petition</a>
-            <a className='icon-link-narrow icon-managepetitions' href='https://petitions.moveon.org/dashboard.html'>Dashboard</a>
+            <Link className='icon-start icon-link-narrow' to='/create_start.html?source=bnav'>Start a Petition</Link>
+            <Link className='icon-link-narrow icon-managepetitions' to='/dashboard.html'>Dashboard</Link>
```
schuyler1d: So these are the good patterns now -- every single link that is to petitions.moveon.org is now a `<Link` and has an absolute url in the `to=` field.

src/components/nav.js
```diff
@@ -30,51 +31,54 @@ const Nav = ({ user, nav, organization }) => {
...
             <div className='moveon-masthead pull-left'>
-              <a href='/'>MoveOn.org</a>
+              <Link to='/'>MoveOn.org</Link>
             </div>
           </div>
-
-          <div className='pull-left top-icons hidden-phone'>
-            <div className='pull-left span2 petitions-partner-logo bump-top-1 margin-right-2 hidden-phone'>
-              {partnerLogoLinks}
+          {!minimal
```
schuyler1d: This is the second part -- basically when we have a minimal property passed in, then we skip all these additional links

src/components/wrapper.js
```diff
@@ -5,9 +5,12 @@ import PropTypes from 'prop-types'
 import Nav from './nav.js'
 import Footer from './footer.js'
 
-const Wrapper = ({ children, params }) => (
+const Wrapper = ({ children, params, routes }) => (
   <div className='moveon-petitions'>
-    <Nav organization={params && params.organization || ''} />
+    <Nav
+      organization={params && params.organization || ''}
+      minimal={!!routes[routes.length - 1].minimalNav}
```
schuyler1d: passes in minimal based on whether the route we're on has a property called `minimalNav` -- an easy decorator to tweak the nav for a route (i think)!

src/routes.js
```diff
@@ -20,28 +20,62 @@ const baseAppPath = process.env.BASE_APP_PATH || '/'
...
+  // that view yet or if we're in production limits matches
+  // to <Route> objects marked with a prodReady property (={true})
+  // All <Link>s and appLocation.push calls in the codebase should NOT be relative
+  // -- i.e. they should be absolute links like /thanks.html
+
+  const PROD_URL = 'https://petitions.moveon.org'
+  const origPush = historyObj.push
+  // eslint-disable-next-line no-param-reassign
+  historyObj.push = (path, state) => {
+    match(
```
schuyler1d: so everything has an ugly side.  This has to overwrite our `historyObj.push` method to test for various situations.  if we pass through, then we just proceed as normal, but if:
* It's a http.... url then we just go to it
* If we're on production (really produciton production -- on petitions) then the route has to be marked `prodReady` or we redirect away from the react app.
* If there's a 404, i.e. no route defined, then we also assume it's for the old site, so we redirect.

src/routes.js
```diff
@@ -20,28 +20,62 @@ const baseAppPath = process.env.BASE_APP_PATH || '/'
...
+  }
+}
+
 export const routes = (store) => {
   const orgLoader = (nextState) => {
     if (nextState.params && nextState.params.organization) {
       store.dispatch(loadOrganization(nextState.params.organization))
     }
   }
+  const routeHierarchy = (
```
schuyler1d: This is just moving the whole hierarchy above the thing we return, since we need the routes to do the appLocation.push test.

webpack.config.js
```diff
@@ -67,7 +67,8 @@ var config = {
         'API_WRITABLE': JSON.stringify(process.env.API_WRITABLE || process.env.PROD),
         'API_SIGN_PETITION': JSON.stringify(process.env.API_SIGN_PETITION || ''),
         'BASE_APP_PATH': JSON.stringify(process.env.BASE_APP_PATH || '/'),
-        'BASE_URL': JSON.stringify(process.env.BASE_URL || 'https://petitions.moveon.org'),
+        'BASE_URL': JSON.stringify(process.env.BASE_URL
```
schuyler1d: Now that we are doing something weird when the BASE_URL = 'petitions.moveon.org' we probably shouldn't set that as the default -- only for PROD (and for stage, we're setting this explicitly)

CONTRIBUTING.md
```diff
@@ -8,5 +8,8 @@
...
+   - The url must be absolute and not relative (start with a `/`)
```
sreynen: I think what you're talking about here is root relative URLs, not absolute. Absolute URLs include https:// and the domain, and that doesn't seem to be the required pattern.

src/routes.js
```diff
@@ -20,28 +20,62 @@ const baseAppPath = process.env.BASE_APP_PATH || '/'
 
 export const appLocation = (Config.USE_HASH_BROWSING ? hashHistory : browserHistory)
 
+const updateHistoryObject = (historyObj, routes) => {
+  // This overrides <Link> routes and router.push() calls if we haven't implemented
+  // that view yet or if we're in production limits matches
+  // to <Route> objects marked with a prodReady property (={true})
+  // All <Link>s and appLocation.push calls in the codebase should NOT be relative
+  // -- i.e. they should be absolute links like /thanks.html
```
sreynen: See note above. Root relative, not absolute.

src/routes.js
```diff
@@ -20,28 +20,62 @@ const baseAppPath = process.env.BASE_APP_PATH || '/'
...
     if (nextState.params && nextState.params.organization) {
       store.dispatch(loadOrganization(nextState.params.organization))
     }
   }
+  const routeHierarchy = (
+    <Route path={baseAppPath} component={Wrapper} onEnter={(nextState) => { store.dispatch(loadSession(nextState)) }} >
+      <IndexRoute component={Home} />
+      <Route path='/sign/:petition_slug' component={SignPetition} />
+      <Route path='/:organization/sign/:petition_slug' component={SignPetition} onEnter={orgLoader} />
+      <Route path='/thanks.html' component={ThanksPage} prodReady minimalNav />
```
sreynen: thanks.html isn't actually prod ready yet. In theory we're only linking to it from sign so this won't matter, but it seems worth entirely avoiding the risk of accidentally deploying something by not adding this in advance.

schuyler1d: I agree it's not prod-ready yet, but I thought it would be useful to have at least one example on the code, and as you noted, it's impossible to get to the thanks page until /sign/ is marked as prodReady.  I can remove it to avoid status-confusion

schuyler1d:
closing for replacement of #224 (needs approval/review again?) and #225 (merged)

PR #218, opened 2018-01-31
Fixes footer to match logged out state in prod

### LINKED ISSUE #206: Update footer links to be logged-out state
Currently looks like this:
screen shot 2018-01-30 at 3 39 24 pm

Should look like this:
screen shot 2018-01-30 at 3 39 17 pm

I.e. it should look how it does when logged out on https://petitions.moveon.org/ not when logged in.

shakalee14:
Closes #206


shakalee14:
I will wait until #200 is merged and then switch to <Link />

sreynen:
#224 (replacement for #200) was merged, so this should be good to merge now after the merge conflict is resolved.

PR #219, opened 2018-01-31
Removes inline styling and switches to Link component

### LINKED ISSUE #208: Font size is too small on "Start a Petition Campaign" button
Looks like it has an inline `font-size: 99%`, perhaps added when font face was wrong.

shakalee14:
Closes #208

### DIFF
--- a/src/components/billboard.js
+++ b/src/components/billboard.js
@@ -1,13 +1,6 @@
 import React from 'react'
 import Video from './video'
-
-// To fix font size issue for red button font in chrome browser
-
-const billboardButton = {
-  fontSize: '99%'
-}
-
-// TODO: make start a petition a <Link> when we have implemented petition creation
+import { Link } from 'react-router'
 
 const BillBoard = () => (
   <div className='container'>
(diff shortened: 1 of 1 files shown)


src/components/billboard.js
```diff
@@ -18,9 +11,9 @@ const BillBoard = () => (
           <div className='billboard-text'>
             <h1 className='white big-title'>People Powered Petitions</h1>
             <p className='white size-medium-large'>There are more than 8 million MoveOn members. Tap into our shared people power and create progressive change.</p>
-            <a href='/create_start.html?source=petitionshomepage' className='button button-xl background-moveon-bright-red' style={billboardButton}>
+            <Link to='/create_start.html?source=petitionshomepage' className='button button-xl background-moveon-bright-red'>
```
sreynen: We're likely to deploy the home page before the create petition page is ready, so we don't want to load create petition page in React just yet.

sreynen:
Now that I'm looking at #200, this might work in that context, so this might only need to wait on that.

PR #222, opened 2018-01-31
212 add search results header text

shakalee14:


shakalee14:
TODO:
- [ ] Instead of rendering state abbrev, state name

PR #224, opened 2018-01-31
Production routing

### LINKED ISSUE #223: distinguish production-ready routes for deployment
This will make it easier to deploy and simultaneously work on components/routes that are not yet fully ready, but still merged on master.

schuyler1d:
fixes #223 

(and same as #200 but without the second commit, and also creates an environment variable if we want to enable this for testing)


src/routes.js
```diff
@@ -19,27 +19,61 @@ const baseAppPath = process.env.BASE_APP_PATH || '/'
...
+  const PROD_URL = 'https://petitions.moveon.org'
+  const origPush = historyObj.push
+  // eslint-disable-next-line no-param-reassign
+  historyObj.push = (path, state) => {
+    match(
+      { location: path, routes },
+      (error, newlocation, props) => {
+        if (!error && props) {
+          const matchedComponent = props.routes[props.routes.length - 1]
+          if (matchedComponent.prodReady || Config.ONLY_PROD_ROUTES || Config.BASE_URL !== PROD_URL) {
```
sreynen: Copying from #200, `|| Config.BASE_URL !== PROD_URL` seems to make it always open everything in React on local, which makes this hard to test, but it also means this won't work on stage, which seems like a problem.

sreynen: I'd missed the new `Config.ONLY_PROD_ROUTES` here, but it doesn't seem to work as described. I think this should be `if (matchedComponent.prodReady || (!Config.ONLY_PROD_ROUTES && Config.BASE_URL !== PROD_URL)) {`?

sreynen: To better describe the problem I see, I've run `export ONLY_PROD_ROUTES=1` and `npm run dev` on my local. I expect the "hot petitions" on the home page to link to petitions.moveon.org, because the sign path isn't marked as `prodReady`, but they still load in React.

schuyler1d:
fixed -- good catch.  do you want to review again?  Feel free to merge, @sreynen as you like

PR #226, opened 2018-01-31
Adds state check box component and dynamic rendering based on state q…

### LINKED ISSUE #217: State search checkbox is missing
When searching by state, the form at the top should have a checkbox to remove the state filter. E.g. /find/?q=science&state=OH should say "[x] Only petitions from Ohio"

shakalee14:
…uery

Currently if you visit http://localhost:8080/#/find/?q=science&state=OH,
It'll show a checkbox

If you visit http://localhost:8080/#/find/?q=science,
It won't show a checkbox

Will fix #217 

TODOS:
- [x] When a user unchecks, make new query without state
NOTE:
On legacy site, after you uncheck state box, search bar aligns left, sometimes?

### DIFF
--- a/src/components/state-check-box.js
+++ b/src/components/state-check-box.js
@@ -1,25 +1,42 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import { connect } from 'react-redux'
 
+import { actions as petitionActions } from '../actions/petitionActions.js'
 import { getStateFullName } from './state-abbrev.js'
+import { appLocation } from '../routes.js'
 
 class StateCheckBox extends React.Component {
 
   constructor(props) {
     super(props)
     this.state = {
-      stateChecked: true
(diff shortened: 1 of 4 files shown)


src/components/state-check-box.js
```diff
@@ -0,0 +1,75 @@
...
+      selectState: props.selectState
+    }
+    this.stateBoxChecked = this.stateBoxChecked.bind(this)
+  }
+
+  componentWillMount() {
+    this.getQueryString()
+  }
+
+  getQueryString() {
```
schuyler1d: too late before merge, but FYI, we shouldn't need to parse query strings in react.  there's a props.location.query object that will already have them parsed that's passed to the page, and you can pass that down to components that need query parameter data.  Above, it's available here already with `props.query`

PR #228, opened 2018-02-01
Forgot password page

### LINKED ISSUE #149: Create forgot password page and component
Page should look like https://petitions.moveon.org/login/forgot_password.html and submit to API, but can be mock API for testing, since response is entirely server-side. After submit, should look like https://petitions.moveon.org/login/forgot_password_thanks.html

jparkrr:
closes #149
api issue: https://github.com/MoveOnOrg/mop/issues/150

Creates http://localhost:8080/#/login/forgot_password.html

For simplicity I have left out:
- All server errors (email is validated locally though)
- Thank you page doesn't have its own route or component
- Redux is not involved

I took a lot of the validation stuff from user registration. I'm not sure if it's worth extracting? https://github.com/MoveOnOrg/mop-frontend/blob/master/src/components/register-form.js

Open to changing anything!

### DIFF
--- a/src/actions/accountActions.js
+++ b/src/actions/accountActions.js
@@ -45,7 +45,7 @@ export function register(fields) {
 
 export function forgotPassword(email) {
   // We don't care much about the response in this case
-  return fetch(`${Config.API_URI}/users/forgotPassword.json`, {
+  return fetch(`${Config.API_URI}/users/forgot-password.json`, {
     method: 'POST',
     body: JSON.stringify({ email })
   })
--- a/src/actions/accountActions.js
+++ b/src/actions/accountActions.js
@@ -43,6 +43,17 @@ export function register(fields) {
   }
 }
 
+export function forgotPassword(email) {
+  // We don't care much about the response in this case
+  return fetch(`${Config.API_URI}/users/forgotPassword.json`, {
+    method: 'POST',
+    body: JSON.stringify({ email })
+  })
+    .then(() => true)
+    .catch(() => true)
+}
+
 export const actions = {
-  register
+  register,
+  forgotPassword
 }
--- a/src/components/forgot-password-form.js
+++ b/src/components/forgot-password-form.js
@@ -1,0 +1,126 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+class ForgotPasswordForm extends React.Component {
+  constructor(props) {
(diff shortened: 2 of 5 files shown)


src/actions/accountActions.js
```diff
@@ -43,6 +43,17 @@ export function register(fields) {
   }
 }
 
+export function forgotPassword(email) {
+  // We don't care much about the response in this case
+  return fetch(`${Config.API_URI}/users/forgotPassword.json`, {
```
sreynen: Our API multi-word endpoints all have dashes between words, so let's do that here, e.g. `forgot-password.json`

src/components/forgot-password-form.js
```diff
@@ -0,0 +1,126 @@
...
+  errorList() {
+    return this.state.presubmitErrors.map((error, idx) => <li key={idx}>{error.message}</li>)
+  }
+
+  renderThanks() {
+    return (
+      <div>
+        An email has been sent containing a link you can use to reset your
+        MoveOn Petitions password.{' '}
+        <b>Thank you for using MoveOn's petition website!</b>
```
sreynen: Straight apostrophes mess up formatting on both GitHub and some of our editors, and curly apostrophes look a little better, so let's use `’` here.

src/components/forgot-password-form.js
```diff
@@ -0,0 +1,126 @@
...
+        >
+          <div className='control-group'>
+            <label className='control-label' htmlFor='inputEmail'>
+              Email:
+            </label>
+            <div className='controls'>
+              <input
+                ref={email => (this.email = email)}
+                name='email'
+                className='percent-80 validation_error'
```
sreynen: Are these classes doing something now that I'm missing, or is this for some future use?

jparkrr: Probably not. I just took them from the live site. I'll remove them if possible

jparkrr: looks like `percent-80` makes gives the input an 80% width, which makes it a bit more balanced IMO. I removed the others and the id. Thanks!

jparkrr:
@sreynen thanks for the review! I've updated it

jparkrr:
@sreynen I added tests here too

sreynen:
Tests look good to me.

PR #229, opened 2018-02-01
User petition overview

### LINKED ISSUE #58: Add petition overview component, add to dashboard page

jparkrr:
Implements the top part of http://localhost:8080/#/dashboard.html

If you have multiple petitions there will be a dropdown to select between them, if you only have one, there will be a "Create petition" button.

When fetching the user's petitions, we add them to the petitionStore, and add their ids in the userStore.petitions. I didn't implement any pagination on the /user/petitions endpoint.

would close #58


jparkrr:
@sreynen can you take a look at this when you have a chance? Not sure if this is in a mergeable state.

sreynen:
Still reviewing, but noting I added #253 as a follow-up task for this.

src/components/legacy/petition-overview.js
```diff
@@ -0,0 +1,104 @@
...
+          <a
+            className='button background-moveon-bright-red'
+            href='mailto:petitions@moveon.org'
+          >
+            Email support for help
+          </a>
+        </div>
+
+        <div className='size-small'>
+          <Link to='/campaign_tips.html' target='_blank'>
```
sreynen: I think anything we're targeting with `<Link>` needs a route in routes.js. That's a little counter-intuitive because we aren't actually loading this in React, but the routes in routes.js are also used to redirect links to load as full URLs outside React.

That said, I'm not sure if this specific path should be added to routes.js, because our current plan is to never add it to React and always load it as a separate HTML document. So maybe this should just be a regular `<a>` link.

It will need a change either way, but I don't yet have a strong preference of which change.

src/components/legacy/petition-overview.js
```diff
@@ -0,0 +1,104 @@
...
+    </div>
+
+    <div className='row'>
+      <div className='span12'>
+        <h3 className='stumpy'>Petition Text</h3>
+
+        <div className='font-lighter lh-24'>{p.summary}</div>
+
+        <div className='size-small'>
+          <Link to={`/edit.html?petition_id=${p.petition_id}`}>
```
sreynen: This one should definitely be added to routes.js. Until it's ready, that will just allow `updateHistoryObject` to see that it needs to be redirected to the legacy URL, so it doesn't matter at all yet which component the route is tied to (could maybe make an empty placeholder component for all such paths), just that the path has a route.

src/components/legacy/petition-overview.js
jparkrr: ok, for now I'll just make it an `<a>`

src/components/legacy/petition-overview.js
jparkrr: It looks to me like it will also be redirected to prod if there's no matching route https://github.com/MoveOnOrg/mop-frontend/blob/a6a0ca0cadf08770ba8b9a7bfc258e953e41be6e/src/routes.js#L35-L51

I can still add placeholders if you think it's better

sreynen:
To loop back here from conversation we had in private chat, I was wrong that `<Link>`s need to be in routes.js, and the only problem is `target='_blank'`, which we can just remove for now.

Issue #148, opened 2018-01-16
Make login form POST login credentials to API, recieve auth cookie


sreynen:
Submitting login form should log you in. Blocked on #147.

jparkrr:
Shouldn't be blocked anymore due to #232

I'll have a look at the API for login, but any pointers are appreciated.

jparkrr:
Looks like the API will be implemented in https://github.com/MoveOnOrg/mop/issues/124

schuyler1d:
This is implemented in branch `auth-phase-2` (https://github.com/MoveOnOrg/mop/pull/287)
I'm not going to merge ot main until we have a branch on mop-frontend that we can test together.  The api points are: `/api/v1/user/session/login` and `logout`.  Login is a json api with `{"username": "email@example.com", "password": "foobar"}` That creates a cookie.

There is still an outstanding CSRF issue for petition creation (and probably other related forms), but login/logout should be doable now.

PR #232, opened 2018-02-01
Add login page (that doesn't work yet)

### LINKED ISSUE #147: Create login page and component without API interaction
Page should look like https://petitions.moveon.org/login/index.html but doesn't need to submit yet.

jparkrr:
closes #147 

Next step would be #148, to make login work.

Again, there was a lot in common with https://github.com/MoveOnOrg/mop-frontend/blob/master/src/components/register-form.js

### DIFF
--- a/src/components/login-form.js
+++ b/src/components/login-form.js
@@ -1,0 +1,131 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Link } from 'react-router'
+
+class LoginForm extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = {
+      presubmitErrors: null
+    }
+    this.handleSubmit = this.handleSubmit.bind(this)
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.errors) {
+      this.setState({ presubmitErrors: null })
+    }
+    this.password.value = ''
+    this.passwordConfirm.value = ''
+  }
+
+  isValidEmail(email) {
+    const regex = /.+@.+\..+/ // Forgiving email regex
+    return regex.test(email)
+  }
+
+  /**
+   * Validates the form for client side errors.
+   * If valid returns true otherwise false.
+   * If errors it will update the local state `presubmitErrors`
+   * @returns {boolean}
+   */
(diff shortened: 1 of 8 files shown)


src/components/login-form.js
```diff
@@ -0,0 +1,131 @@
...
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.errors) {
+      this.setState({ presubmitErrors: null })
+    }
+    this.password.value = ''
+    this.passwordConfirm.value = ''
+  }
+
+  isValidEmail(email) {
```
sreynen: Let's move this to lib.js since we're using it in a few places now.

src/routes.js
```diff
@@ -41,6 +42,7 @@ export const routes = (store) => {
         <Route path='/:organization/create_start.html' component={CreatePetitionPage} onEnter={orgLoader} />
         <Route path='/:organization/' component={Home} onEnter={orgLoader} />
         <Route path='/login/register.html' component={Register} />
+        <Route path='/login/index.html' component={Login} />
```
sreynen: This should also respond to just `/login/`, as the current version does: https://petitions.moveon.org/login/ That might conflict with `/:organization/`. If so, we can actually remove `/:organization/`, as we recently decided it's no longer needed.

jparkrr: Good point! I fixed the conflict by moving `/:organization/` after the new `/login/`.

jparkrr:
Thanks for the review! Should be good now.

test/components/login-form.js
```diff
@@ -0,0 +1,49 @@
...
+import LoginForm from '../../src/components/login-form'
+
+describe('<LoginForm />', () => {
+  it('displays fields', () => {
+    const login = shallow(<LoginForm />)
+    expect(login.find('input[name="email"]').length).to.equal(1)
+    expect(login.find('input[name="password"]').length).to.equal(1)
+  })
+
+  it.skip('does something different when logged in (TODO)', () => {})
```
sreynen: Does it actually do something different when logged in? https://petitions.moveon.org/login/ doesn't seem to.

PR #239, opened 2018-02-02
 Sample data for dashboard APIs

franciskung:


jparkrr:
Do you think it would be a good idea to add a second user petition? Definitely won't be the case for all users, but something that needs to be handled.

jparkrr:
I also have a question related to the user having multiple petitions. In the current version, when you go to http://petitions.moveon.org/dashboard.html, one of your petitions is displayed, even if you have multiple. (there's also a dropdown to switch)

How do we decide which to display by default? Can we just use the first returned by `/user/petitions`?

franciskung:
Yes, they are sorted by timestamp (descending) so the most recent one will always be first.  I can add another petition or two, to demo multiple petitions in that list.

We debated whether to have an endpoint like, say, /user/petitions/latest.json that only gave you details on the latest petition, and then require frontend to do a second call to /user/petitions.json for the other petitions in the dropdown... but figured it's probably eaiser to just return a bunch and display the first (plus, the others can get cached by the client)

However I'm open to changing this around if something else is easier for you.

Issue #235, opened 2018-02-02
query-string should not be a dependency


schuyler1d:
remove query-string from
 src/components/state-check-box.js
(which breaks tests when I run it)
and instead of parsing query string, we should use the `props.query` parameter in that component already passed

6ewis:
I couldn't reproduce the test breaking - it ran fine with query-string. However, I updated it to use props.query https://github.com/MoveOnOrg/mop-frontend/pull/252 .

PR #252, opened 2018-02-07
fix - query-string should not be a dependency #235

6ewis:


sreynen:
This is great, fixes 3 issues with less code. Thanks!

PR #259, opened 2018-02-07
implements page view tracking for google/segment (#255)

schuyler1d:
Implements fix for #255

### DIFF
--- a/src/actions/sessionActions.js
+++ b/src/actions/sessionActions.js
@@ -33,6 +33,15 @@ function callSessionApi(tokens) {
           session: json,
           tokens
         })
+        // segment tracking
+        if (window.analytics && window.analytics.identify
+            && json && json.identifiers && json.identifiers.length) {
+          json.identifiers.forEach((id) => {
+            if (/^actionkit:/.test(id)) {
+              window.analytics.identify(id.substring('actionkit:'.length))
+            }
+          })
+        }
       }),
       (err) => {
         dispatch({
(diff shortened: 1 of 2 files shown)


src/actions/sessionActions.js
```diff
@@ -71,6 +80,24 @@ export const loadSession = ({ location }) => {
   }
 }
 
+export const trackPage = (newLocation) => {
+  // we track at POP -- at PUSH, we haven't arrived yet!
+  if (newLocation.action === 'POP') {
+    if (window.analytics) {
+      // Segment.com tracking
+      // https://segment.com/docs/sources/website/analytics.js/#page
```
schuyler1d: I did this subconsciously, but I totally expect segment's documentation urls to persist longer than google's which change every 6 months!

src/actions/sessionActions.js
```diff
@@ -71,6 +80,24 @@ export const loadSession = ({ location }) => {
...
+      window.analytics.page({
+        // overrides any canonical url set
+        location: document.location.href,
```
sreynen: The current site and backend and passing in `sanitized_location= String(document.location).replace(/\?source=/, '?utm_source=').replace(/\&source=/, '&utm_source=');` and we're using ?source= in at least one location: https://github.com/MoveOnOrg/mop-frontend/blob/master/src/components/legacy/petition-preview.js#L8

I'm not sure if we should change that to utm_source in the React context, or if this needs to "sanitize" but guessing we want to do one of those.

schuyler1d: I'm going to wait for @sandramchung to weigh in -- it seemed like she had an idea that we could do this by config rather than code-hacking the url.  If so, once we determine that to be true, let's document :-)

Issue #203, opened 2018-01-30
Validate and submit petition create form step 1


sreynen:
Validation can probably copy a lot from how signature form does validation. When validation succeeds, the form values should save in redux and then move to step 2, which is covered in #339.

jparkrr:
@shakalee14 I have a refactor of the petition create stuff in https://github.com/MoveOnOrg/mop-frontend/pull/279. Sorry that it might conflict with the things you have locally for this. Let me know if you want me to take this over.

jparkrr:
Further work depends on https://github.com/MoveOnOrg/mop/issues/284

PR #260, opened 2018-02-08
WIP: Petition target autocomplete

jparkrr:
Still need to work on the design a little bit. I need to figure out what to do with the css, because it doesn't seem present in the css file that we're already loading.

Currently I don't use the `search` parameter in the api at all, just load the entire `national` or `state` group onFocus and do filtering on the client. Part of the reason for this is the API seems pretty slow. But I think this new approach seems better in general, since the json would only be a few hundred entries at most.

I added a new env variable for the legacy api, since it resides outside `API_URI`.


jparkrr:
Kind of depends on https://github.com/MoveOnOrg/mop-frontend/issues/203, but we're working on dividing that into separate PRs. cc @shakalee14

jparkrr:
Styling is basically matching now. On the old version it came from jquery-ui.

So far I kept it as inline react style, which I think is fine since it's likely to be redesigned at some point, but let me know if you think I should copy it to a CSS file or something :)

PR #261, opened 2018-02-09
WIP: Component switch w/ webpack

### LINKED ISSUE #256: Make home page example component set switch
We've started moving some components to src/components/legacy in anticipation of having new versions of those same components with new designs that use slightly different HTML but mostly function the same. The task here is to work out how we're going to switch between these component sets, using the home page as an example.

Ideally we'd be able to change a single variable (maybe a parameter on the route?) and make everything on the home page start using components from src/components/giraffe instead of src/components/legacy. It would also be great if we could keep the initial load size smaller by lazy-loading those components, so we're not loading components from both sets when we're only using one set.

jparkrr:
WIP to fix #256, but I would love comments.

We have a lot of things now, some of which might not be necessary, but I wanted to try it out

- `npm build`
- `npm build-legacy`
- `npm dev-build`
- `npm dev-build-legacy`
- `npm dev`
- `npm dev-legacy`

All those with `-legacy` replace as follows:
```
import BillBoard from 'Giraffe/billboard' -> import BillBoard from 'src/components/legacy/Giraffe/billboard'
```

For consistency, I added `Legacy` as an alias as well, so if you want a component to always be legacy, you can do:
```
import BillBoard from 'Legacy/billboard' 
```


.eslintrc.js
```diff
@@ -13,6 +13,7 @@ module.exports = {
     'no-alert': ['off'],
     'jsx-quotes': ['error', 'prefer-single'],
     'max-len': ['off'],
-    'prefer-object-spread/prefer-object-spread': 2
+    'prefer-object-spread/prefer-object-spread': 2,
+    'import/no-unresolved': 'off'
```
jparkrr: You can supposedly make this thing work, so we don't have to turn it off https://www.npmjs.com/package/eslint-import-resolver-webpack

package.json
```diff
@@ -2,16 +2,19 @@
...
   "directories": {},
   "scripts": {
-    "test": "mocha --compilers js:babel-core/register --require jsdom-global/register --require isomorphic-fetch test test/*",
-    "test1": "mocha --compilers js:babel-core/register --require jsdom-global/register --require isomorphic-fetch test ",
-    "build": "webpack",
-    "dev-build": "webpack -d",
-    "webpack-watch": "webpack -w",
-    "dev": "node ./node_modules/webpack-dev-server/bin/webpack-dev-server.js --hot --progress --colors --port ${PORT:-8080} --inline",
+    "test": "NODE_ENV=test mocha --compilers js:babel-core/register --require jsdom-global/register --require isomorphic-fetch test test/*",
+    "test1": "NODE_ENV=test mocha --compilers js:babel-core/register --require jsdom-global/register --require isomorphic-fetch test ",
```
jparkrr: do we need `test1`?

package.json
```diff
@@ -2,16 +2,19 @@
...
+    "build": "webpack --config webpack/base.config.js",
+    "build-legacy": "webpack --config webpack/legacy.config.js",
+    "dev-build": "webpack -d --config webpack/base.config.js",
+    "dev-build-legacy": "webpack -d --config webpack/legacy.config.js",
+    "webpack-watch": "webpack -w --config webpack/base.config.js",
```
jparkrr: Do we still want `Webpack-watch`? Might be nice to clean it up if its not used

package.json
schuyler1d: test1 is useful to test a single file, which was helpful for me when getting a specific test to work.

schuyler1d:
Also, I think we need this line:
https://github.com/MoveOnOrg/mop-frontend/blob/66dd4fabdbc9ac2a090b627c0da90822a7aff1fd/webpack/base.config.js#L22
(along with maybe a bunch of other places) to output a name that's also based on process.env.STYLE since we want to be able to deploy these built files alongside each other

jparkrr:
Thanks for the review @schuyler1d. I'll change it to how you mention in a new PR.

Your way is definately more consistent with the current use of env vars, as well as a simpler diff, which is nice! It doesn't match things I've seen out there before with webpack, but consistency is more important.

The only concern I have is having to onboard people as to what the `Style` alias means (I think it's more explicit if it matches a folder). Though, it will probably only be a small stumbling block that they can hopefully figure out from context and/or documentation.

PR #264, opened 2018-02-10
Redirect to/from org sign URLs to match petition org

### LINKED ISSUE #193: mis-located megapartner url should redirect in-app to be without the megapartner url component
See backend issue here: https://github.com/MoveOnOrg/mop/issues/137

Basically, the react app should be responsible for "redirecting" to a url of the form: `/<megapartner>/sign/<non-megapartner-petition>` to `/sign/<non-megapartner-petition>` while preserving query parameters.

This way the backend doesn't need to be responsible for handling all redirects of a mis-located petition.

sreynen:
Fixes #193.


src/pages/sign-petition.js
```diff
@@ -6,12 +6,38 @@ import { connect } from 'react-redux'
...
 import { actions as petitionActions } from '../actions/petitionActions.js'
+import { appLocation } from '../routes.js'
 
 class SignPetition extends React.Component {
 
   componentWillMount() {
-    const { dispatch, params } = this.props
+    const { dispatch, params, petition } = this.props
     dispatch(petitionActions.loadPetition(params.petition_slug))
+    if (petition) {
```
sreynen: This happens if the petition is loaded before the component.

src/pages/sign-petition.js
```diff
@@ -6,12 +6,38 @@ import { connect } from 'react-redux'
...
+      this.checkOrgPathMatches(petition, params.organization)
+    }
+  }
+
+  componentWillReceiveProps(nextProps) {
+    const { params, petition } = nextProps
+    if (petition) {
```
sreynen: This happens if the petition loads after the component.

PR #266, opened 2018-02-12
Component switch w/ webpack

### LINKED ISSUE #256: Make home page example component set switch

jparkrr:
New PR based on @schuyler1d's feedback in #261

WIP to fix #256

You can set an env variable `THEME` as `giraffe` or `legacy` (default) for `npm dev` or `build`.

If `THEME=giraffe`, the alias will be rewritten like this:
```
import Billboard from 'Theme/billboard' -> import Billboard from 'src/components/theme-giraffe/billboard'
```

For consistency, I added `LegacyTheme` as an alias as well, so if you want a component to always be legacy (because it hasn't been built for giraffe yet), you can do:
```
import Billboard from 'LegacyTheme/billboard' 
```


jparkrr:
@schuyler1d I tried implementing resolution fallback with https://github.com/kije/webpack-fallback-directory-resolver-plugin. It worked, but the only issue was it was required to specify the `.js` extension. I could probably fork and fix that, but it seems like that would be introducing a lot of brittleness and confusion.

jparkrr:
Didn't realize before that we were already doing code splitting here https://github.com/MoveOnOrg/mop-frontend/blob/master/src/loaders/petition.js. I've updated it so the chunks are distinct per theme too.

I've also tested building now, and everything seems to work.

PR #269, opened 2018-02-12
Refactor SignPetition

jparkrr:
Works with mock data, I'll try to test it against local mop or stage api before merging.


src/components/theme-legacy/signature-add-form.js
```diff
@@ -0,0 +1,340 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Link } from 'react-router'
+
+import CountrySelect from '../form/country-select'
+import StateOrRegionInput from '../form/state-or-region-input'
+import ZipOrPostalInput from '../form/zip-or-postal-input'
+
+const SignatureAddForm = ({
```
jparkrr: I wonder if I should try to split this up more as part of this refactor

jparkrr: I'm thinking it's not worth the time. Whoever implements the rebrand could still create smaller components in the giraffe version if they wanted.

src/components/theme-legacy/signature-add-form.js
```diff
@@ -0,0 +1,340 @@
...
+            and MoveOn.org. By signing, you agree to receive email messages from{' '}
+            <span id='organization_receive'>{creator.organization}, </span>MoveOn
+            Political Action, and MoveOn Civic Action. You may unsubscribe at
+            any time. [<Link to='/privacy.html'>privacy policy</Link>]
+          </p>
+        ) : (
+          <p className='disclaimer bump-top-1'>
+            <b>Note:</b> By signing, you agree to receive email messages from
+            MoveOn.org Civic Action and MoveOn.org Political Action. You may
+            unsubscribe at any time. [{' '}
```
jparkrr: sorry about all these weird `{' '}` things. I use https://prettier.io and it adds them. I guess they make sense? I'll think more about it.

jparkrr: I cleaned them up as best I could. You need them when you have two inline elements on different lines.

jparkrr:
AFAICT this is good to go. It seems hard to test everything, but for viewing petitions I manually toggled all of the options and things like that.

src/components/theme-legacy/signature-add-form.js
sreynen: Yeah, I think the rebrand is the right time to split this up more, as that will also have some context of which components might be reusable on other sites in the future.

test/components/signature-list-item.js
```diff
@@ -14,25 +14,25 @@ describe('<SignatureListItem />', () => {
   const createdDate = String(new Date(1836, 11, 10))
 
   it('is a div.signer', () => {
-    const context = shallow(<SignatureListItem user={user} createdDate={createdDate} number={1000} />)
-    expect(context.name()).to.equal('li')
+    const context = mount(<SignatureListItem user={user} createdDate={createdDate} number={1000} />)
```
sreynen: This doesn't need to happen now, but as we're hitting Travis timeouts, these could probably target the component rather than the container (since they all seem to be only testing render) and then stick to `shallow`.

jparkrr: Yes, there's a lot of test refactoring that should happen. I have a good plan for it

1. test the props the child is called with in the container
2. Store those props somewhere and pass to the stateless component and run display tests)

I've been struggling a bit with the redux part though. I'll take some more time to look at it soon.

BTW, this wasn't the test that was timing out. Definitely ideally would be shallow though.

PR #270, opened 2018-02-13
Refactor ThanksPage

jparkrr:
Here we refactor to add a component `loaders/ThanksShim` (previously ThanksPage) that is used in place of the `Thanks` in the router, which handles lazy-loading the actual Thanks component.

Tested with http://localhost:8080/#/thanks.html?petition_id=95983&name=outkast&from_source=homepage, and it seems to work.


src/containers/thanks-loader.js
```diff
@@ -5,7 +5,7 @@ import { connect } from 'react-redux'
 import { thanksLoader } from '../loaders/petition.js'
 import { actions as petitionActions } from '../actions/petitionActions.js'
 
-class ThanksPage extends React.Component {
+class ThanksLoader extends React.Component {
```
sreynen: This file contains another variable, `thanksLoader`, with almost the same name. I think we need more distinct names for those two things. I would argue the container isn't itself a loader, it just calls the loader. Since that's all this does and we already have another container named "Thanks", seems like this should probably merge with that.

jparkrr: The issue is that the "ThanksLoader" (which I agree isn't a good name) requires the other "Thanks" component asynchronously (so that webpack splits it into a chunk).

Maybe something like ThanksShim?

Or maybe the component that has the loader can be in another folder for components that's only job are to require asynchronously.

I'll play around with it and see if I can find a good option.

jparkrr:
@sreynen `Thanks` for the review! I updated the PR description with the changes I just made.

jparkrr:
Let me know if you can think of a better name than `ThanksShim`

PR #285, opened 2018-02-15
Quick fix to remove padding under the video

### LINKED ISSUE #284: Remove padding under video
Currently the following contexts have some padding under the video:

- broom.moveon.org in Firefox
- broom.moveon.org in Chrome
- broom.moveon.org in Safari
- petitions.moveon.org in Firefox
- petitions.moveon.org in Safari

And this context has no such padding:

- petitions.moveon.org in Chrome

So this is technically a regression, but only in Chrome. Ideally we could go beyond fixing it in Chrome and remove the padding in all browsers.

jparkrr:
Quick fix to remove padding under the video as mentioned in #284. I tested in chrome, safari, and Firefox.

I guess the root cause is that the `clearfix::after` adds `content: ""`.

A better fix is probably finding a way to do it without the clearfix.

Also uses ResponsiveEmbed so it works on mobile

(fixes #284)


jparkrr:
Also moves video component to `component/theme-legacy`, which I forgot to do before.

jparkrr:
@sreynen Nice! Updated

jparkrr:
Note that this was indeed updated to work on smaller screens

PR #289, opened 2018-02-16
Ability to toggle nav on mobile

jparkrr:


sreynen:
Retroactively created an issue for this, just so we have a clearer record of what happened: https://github.com/MoveOnOrg/mop-frontend/issues/290

src/components/theme-legacy/nav.js
```diff
@@ -1,15 +1,21 @@
...
 import { Link } from 'react-router'
+import classNames from 'classnames'
 
 import NavLink from 'LegacyTheme/nav-link'
 
-const Nav = ({ user, nav, organization, minimal }) => {
+const Nav = ({ user, nav, organization, minimal, toggleOpen, isOpenMobile }) => {
   const cobrand = ((organization) ? nav.orgs[organization] : nav.partnerCobrand)
 
+  const ulClassNames = classNames({
```
sreynen: I'm not enthusiastic about this new dependency. Seems like it doesn't really buy us much over doing the logic more directly, which I think would be something like this:
```
const ulClassNames = `nav ${isOpenMobile ? '' : 'collapse nav-collapse'}` 
```
But maybe that's just me, worth getting a third opinion here.

jparkrr: I think we will also want to use it with the rebrand, because it's BEM syntax and that can get even more annoying.

sreynen: Alright, I'm sold.

sandramchung: ++ classNames

jparkrr: Yay! I didn't want to bias you, but it's one of my favorite libraries. Mostly from back when I wrote CoffeeScript

with coffee script you could just write

```
classNames nav, 'collapse nav-collapse': isOpen
```
and it would automatically pass an object to the function and everything would be awesome

PR #291, opened 2018-02-19
Giraffe nav implementation and components 

jparkrr:
In implementing the giraffe header/nav, I created a component library in `/src/giraffe-ui` .
This is meant to be good for readability inside this project and could also easily be ported to another react project.

To get a sense of how it can be used, look at the new implementation of [theme-giraffe/nav.js](https://github.com/MoveOnOrg/mop-frontend/blob/13e5438ac51ca73591cfe5f4c176bf15535bac45/src/components/theme-giraffe/nav.js)
compared to the legacy theme [theme-legacy/nav.js](https://github.com/MoveOnOrg/mop-frontend/blob/13e5438ac51ca73591cfe5f4c176bf15535bac45/src/components/theme-legacy/nav.js)

To run, `THEME=giraffe npm run dev`


jparkrr:
Some potential arguments against componentization: it could take more time, and it maybe introduces too many levels of indirection to inhibit understanding.

src/containers/wrapper.js
```diff
@@ -0,0 +1,40 @@
...
+import PropTypes from 'prop-types'
+import Nav from './nav'
+import Footer from 'LegacyTheme/footer.js'
+
+import Config from '../config'
+
+class Wrapper extends React.Component {
+  componentWillMount() {
+    if (Config.THEME === 'giraffe') {
+      document.documentElement.className = 'giraffe'
```
jparkrr: Not sure if this is a good idea yet

package.json
```diff
@@ -64,6 +64,7 @@
     "redux-mock-store": "^1.2.3",
     "redux-test-utils": "^0.2.2",
     "sinon": "^4.2.2",
+    "svg-react-loader": "^0.4.5",
```
sreynen: Looks like this isn't being used yet, so maybe I'm just not seeing the full value of it yet, but I'm initially skeptical this is worth using. Seems easier to just make the SVGs components directly, e.g. `export default const Close = () => (<svg ...>)` and then we can import it without a module. The main downside I see of making them normal components is we can no longer load them as SVGs directly, but I don't expect we'll be doing that anyway, at least not from this repo.

sreynen:
I'm generally in favor of smaller components, but I don't see the point of the extra "components" directory. E.g. I think `src/components/theme-giraffe/components/nav/header.js` should just be `src/components/theme-giraffe/nav/header.js`

jparkrr:
@sreynen my idea is to separate "reusable" components from their implementation in the petition app. Does that make sense?

jparkrr:
I think it especially makes sense to separate components from their implementation if we want an easy possibility of extracting a component library to reuse across react apps.

These components, like `nav/header.js`, don't necessarily have to live within `theme-giraffe/`, just their implementation in `theme-giraffe/nav.js`. So my first proposal is to put them somewhere like `/src/giraffe-components`.

A separate second proposal is having the reusable components in `theme-giraffe/ui/`, equally colocated, but inside the giraffe folder.

Or we can not worry too much about the component library for now and just go ahead and mix things in `giraffe-theme`. One argument for this is probably won't be too hard to extract components to another app later, when that time comes.

src/containers/wrapper.js
schuyler1d: I would much rather that this be just in our own top page template rather than editing the actual top document element.

local/index.html
```diff
@@ -7,6 +7,8 @@
     <meta name="viewport" content="width=device-width, initial-scale=0.9" />
     <link rel='stylesheet' href='//fonts.googleapis.com/css?family=Source+Sans+Pro%3A200%2C300%2C400%2C600%2C700&#038;ver=3.6' type='text/css' media='all' />
     <link rel='stylesheet' href="<%= htmlWebpackPlugin.options.cssPath %>" type='text/css' media='all' />
+    <link rel='stylesheet' href="http://moveon-dev.netlify.com/styles/vendors.css" type='text/css' media='all' />
```
schuyler1d: I would rather this somehow be connected to cssPath in webpack, e.g. here:
https://github.com/MoveOnOrg/mop-frontend/blob/master/webpack.config.js#L58
Maybe we need that to be a path and not the full filename, too, and then you can do giraffe-vendors.css and giraffe-main.css or something. @sandramchung 
Basically, I have a couple of use cases:
1. I want to be able to do this offline, if I have local copies of these (or to be able to build new css versions, and test them locally)
2. let's start getting `giraffe-` in the css filenames, because `main.css` will be a very scary thing to link to if we don't know 'which main.css' it is.
3. Maybe we can't depend on all these static files being in the same place (old and new), but I think that would be useful enough.

src/components/theme-giraffe/components/nav/logo.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import { Link } from 'react-router'
+export const Logo = () => (
+  <Link to='/' className='logo'>
+    <img
+      src='http://moveon-dev.netlify.com/images/logo.svg'
```
schuyler1d: no hard-coding urls and definitely not to netlify.com.  Also, we can't have any non-https urls -- everyhting should be https (default/hard-coded is ok) This should probably be based on STATIC_ROOT environment variable, though I see we haven't been doing that (we should going forward though)

src/components/theme-giraffe/components/nav/primary.js
```diff
@@ -0,0 +1,50 @@
...
+import PropTypes from 'prop-types'
+import cn from 'classnames'
+
+import CaretDownSvg from '../../svgs/caret-down.svg'
+
+const baseCn = 'mo-nav__primary'
+export const Primary = ({ openSections, toggleSection, children }) => (
+  <ul className={baseCn}>
+    {React.Children.map(children, child =>
+      React.cloneElement(child, { openSections, toggleSection })
```
schuyler1d: what is this doing?  Can you add a comment here at least?  Is this kind of meta-programming necessary here (maybe it is, but please explain why)?

src/components/theme-giraffe/components/nav/primary.js
```diff
@@ -0,0 +1,50 @@
...
+
+Primary.propTypes = {
+  openSections: PropTypes.array.isRequired,
+  toggleSection: PropTypes.func.isRequired,
+  children: PropTypes.node
+}
+
+const Section = ({ openSections, toggleSection, topLevel, name, children }) => {
+  const dropdownCn = section =>
+    cn('mo-nav__dropdown', {
```
schuyler1d: this and the below classname abstraction stuff seems useful, but it seems like it should be centralized (maybe in theme-giraffe/style-lib.js or something like that?).  Even more ideally, to someplace that also abstracts these pieces in the same way our styleguide does.

src/components/theme-giraffe/svgs/burger.svg
```diff
@@ -0,0 +1 @@
+<svg width="18" height="12" xmlns="http://www.w3.org/2000/svg"><path d="M0 12h18v-2H0v2zm0-5h18V5H0v2zm0-7v2h18V0H0z"/></svg>
```
schuyler1d: I bet we need these hard-coded for react here, but  these are part of the style, right?  How do we balance the need to centrally reference all these little style/svg files across sites? @sandramchung 

Can/should we at least put all the duplicated assets somewhere in one directory?

jparkrr: It's the css that requires that they be loaded from the same domain, since we apply styles to the svgs https://css-tricks.com/svg-use-with-external-reference-take-2/#comment-1597163

I'm not sure what you mean about duplicated assets

src/components/theme-giraffe/components/nav/primary.js
jparkrr: You have to do this to apply props to children. Good point that I should add a comment

src/components/theme-giraffe/svgs/burger.svg
schuyler1d: wow, TIL about CORS/SVG/CSS -- insane.  I mean across multiple properties.  They won't be duplicated in this project, but they'll be duplicated across probably many of our site's separate repos, possibly.

src/components/theme-giraffe/components/nav/primary.js
jparkrr: Maybe I'll return to this once I've implemented more elements and see what is useful in a library. I can't quite visualize it right now.

src/containers/wrapper.js
jparkrr: With https://github.com/MoveOnOrg/giraffe/pull/15 and https://github.com/MoveOnOrg/mop-frontend/pull/291/commits/06cc0db35e40835659eb20197c2e2b0b2a74421c this is now unnecessary.

local/index.html
jparkrr: Very good point. See https://github.com/MoveOnOrg/mop-frontend/pull/291/commits/0bb3dc33c44d6927932ef9e8eaa49ecf4c5d0c77 for a new way to work offline and develop css.

src/components/theme-giraffe/components/nav/logo.js
jparkrr: Good idea with STATIC_ROOT. I implemented it for dev and stage here https://github.com/MoveOnOrg/mop-frontend/pull/291/commits/2ec10f4a028d5013c6cd2acd80fc6d6a6d2ebb54

jparkrr:
Thank you @schuyler1d. I addressed much of your feedback.

As I proposed yesterday, I moved the giraffe UI elements to `/src/giraffe-ui` in https://github.com/MoveOnOrg/mop-frontend/pull/291/commits/34b2791514fb41bc2132d6bad7394953e7aaec3e.

Now I'm going to look quickly into tests, as right now they aren't running in the giraffe context.

webpack.config.js
```diff
@@ -58,6 +63,12 @@ var config = {
       cssPath: (process.env.NODE_ENV == 'production'
                 ? 'https://s3.amazonaws.com/mop-static/css/moui.css'
                 : '/css/moui.css'),
+      giraffeCssPath: (process.env.LOCAL_CSS
+        ? 'http://localhost:3000/styles/main.css'
+        : 'http://s3.us-east-2.amazonaws.com/static.moveon.org/giraffe/styles/main.css'),
```
schuyler1d: these need to be https links

jparkrr:
Everything seems to be working fine for both
`npm run dev` (normal version) and `THEME=giraffe npm run dev` in development.

This will need https://github.com/MoveOnOrg/mop/pull/192 to work on the staging server, but we don't necessarily need that to merge since, on stage, you have to specifically request a giraffe page to load giraffe components.

webpack.config.js
```diff
@@ -58,6 +63,12 @@ var config = {
       cssPath: (process.env.NODE_ENV == 'production'
                 ? 'https://s3.amazonaws.com/mop-static/css/moui.css'
                 : '/css/moui.css'),
+      giraffeCssPath: (process.env.LOCAL_CSS
+        ? 'http://localhost:3000/styles/main.css'
+        : 'https://s3.us-east-2.amazonaws.com/static.moveon.org/giraffe/styles/main.css'),
+      giraffeCssVendorsPath: (process.env.LOCAL_CSS
+        ? 'http://localhost:3000/styles/vendors.css'
+        : 'https://s3.us-east-2.amazonaws.com/static.moveon.org/giraffe/styles/vendors.css'),
```
schuyler1d: Should we reimplement the css/js for any carossel on the petition homepage? And if there isnt a carosel on the petition homepage maybe we dont need this file

jparkrr: Good point. I'm guessing we don't need it.

jparkrr:
@schuyler1d I added a note to docs/REFERENCE--environment_variables.md, but it's easy to miss
https://github.com/MoveOnOrg/mop-frontend/pull/291/files#diff-3e505f60f9f932445517867ed8e7b7b0R16

It uses port 3000, so it's meant to be served from `gulp watch` of the giraffe repo. I agree it isn't the best to require (1) have the giraffe repo and (2) run the watch process, but I thought it was better than having outdated css. It is also useful for developing the css.

We can add a third option for serving from `/local` if you want

PR #293, opened 2018-02-21
Setup for giraffe tests

jparkrr:
Does two things:

1. Makes travis run tests twice, once with the legacy theme, once with giraffe theme. It's kind of annoying to have to run tests twice, but I don't see a way around it. If we want to build two javascript files, we should run tests for both.

2. Uses `require-hacker` to make importing SVGs work by replacing them with null components in tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ node_js:
 - '6.1'
 script:
   - npm run lint
-  - npm run test
+  - THEME=legacy npm run test
   - THEME=giraffe npm run test
 notifications:
   slack:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -1,7 +1,10 @@
 language: node_js
 node_js:
 - '6.1'
-script: npm run lint-and-test
+script:
+  - npm run lint
+  - npm run test
+  - THEME=giraffe npm run test
```
schuyler1d: is it worth being explicity on the null case here? `THEME=legacy npm run test`
That way if we switch the default we'll still be running on both.  Maybe we should do the same with the `build` command.

PR #296, opened 2018-02-22
move travis stage target to match our naming scheme

schuyler1d:
This PR is a public service announcement to the 'new regime' on static deployment.

Basically giraffe, mop static files, mop-frontend (along with css/moui.css) need to be deployed to different places on mop-static-stage and on production they will go to the same places on static.moveon.org s3 bucket.

We will need to retire mop-static s3 bucket at some point.

### DIFF
--- a/src/routes.js
+++ b/src/routes.js
@@ -64,7 +64,7 @@ export const routes = (store) => {
     }
   }
   const routeHierarchy = (
-    <Route path={baseAppPath} component={Wrapper} onEnter={(nextState) => { store.dispatch(loadSession(nextState)) }} >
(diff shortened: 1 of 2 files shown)


src/routes.js
```diff
@@ -64,7 +64,7 @@ export const routes = (store) => {
     }
   }
   const routeHierarchy = (
-    <Route path={baseAppPath} component={Wrapper} onEnter={(nextState) => { store.dispatch(loadSession(nextState)) }} >
+    <Route path={window.baseAppPath || baseAppPath} component={Wrapper} onEnter={(nextState) => { store.dispatch(loadSession(nextState)) }} >
```
schuyler1d: This is so we can do different urls

PR #299, opened 2018-02-23
Petition reducer tests

lady3bean:


src/reducers/index.js
```diff
@@ -64,7 +64,7 @@ function petitionReducer(state = initialPetitionState, action) {
   let petition = {}
   let updateData = {}
   if (typeof petitionWithoutSlug === 'object') {
-    petition = Object.assign(petitionWithoutSlug, { slug })
+    Object.assign(petition, petitionWithoutSlug, { slug })
```
jparkrr: Very good find!

You might consider writing this as the following, as people often do to indicate that a new object is being created (even though we can see it a couple of lines above in this case)
`petition = Object.assign({}, petitionWithoutSlug, { slug })`

but then we have an eslint rule that tells us that we should "prefer object spread", so it would look like this:
`petition = { ...petition, ...petitionWithoutSlug, ...{ slug } }`

PR #302, opened 2018-02-23
Petition reducer tests

sreynen:
Repeat of https://github.com/MoveOnOrg/mop-frontend/pull/299 against new default branch


sreynen:
@lady3bean I had to recreate this PR after I switched the default branch, and now it shows up as my PR so I can't approve it the normal way, but 👍

PR #305, opened 2018-02-26
Giraffe sign petition (part 1)

jparkrr:
Trying to divide into multiple PRs.

This one contains the petition card (the blue box). It was a little unclear from the design, so I just put the petition summary here. These parts of the petition can easily be moved around.

It also contains the petition details under the card contains the full text of the petition as well as the author and some share buttons that aren't hooked up yet.

It also contains the signature add form that actually works pretty well, thanks to the logic that already existed. (Small design changes needed). I was able to edit all of the existing tests for this component to apply equally to both the legacy and giraffe versions.

Tracking issue: #309


src/lib/index.js
```diff
@@ -40,6 +40,13 @@ export const text2paraJsx = (str) => {
   })
 }
 
+// Helps the css work properly
+export const splitIntoSpansJsx = (str) => (
+  str
+  .match(/[\S]+/gi)
+  .map((word, i) => <span key={i}>{word}</span>)
```
jparkrr: This is kind of a weird thing to do to our petition headings, but it comes straight from the design.

src/giraffe-ui/petition/country-select.js
```diff
@@ -0,0 +1,246 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import CaretDownSvg from '../svgs/caret-down.svg'
+import cx from 'classnames'
+
+export const CountrySelect = ({ value, onChange, className }) => (
+  <div className={cx('input-block', className, { active: !!value })}>
+    <select name='country' id='country' value={value} onChange={onChange}>
+      <option>United States</option>
```
jparkrr: These country options aren't currently shared with legacy. Maybe I should put them in ./lib?

schuyler1d: I vote yes on that

src/giraffe-ui/petition/state-select.js
```diff
@@ -0,0 +1,86 @@
...
+import PropTypes from 'prop-types'
+import cx from 'classnames'
+
+import CaretDownSvg from '../svgs/caret-down.svg'
+
+export const StateSelect = ({ value, onChange, className }) => (
+  <div className={cx('input-block', className, { active: !!value })}>
+    <select name='state' id='state' className={className} onChange={onChange}>
+      <option></option>
+      <option value='AL'>Alabama</option>
```
schuyler1d: Can the state list also be shared across themes? It can just be the value/display pairs and then render options can even be diff

src/giraffe-ui/petition/share.js
```diff
@@ -0,0 +1,52 @@
...
+  return (
+    <div className={`${className}__share`}>
+      <strong className={`${className}__share-heading`}>Share</strong>
+      <div className={`${className}__share-${hasLabels ? 'items' : 'icons'}`}>
+        <Link
+          to='#'
+          onClick={mail}
+          className={cx(elementCn, `${elementCn}--mail`)}
+        >
+          <svg><use xlinkHref='#mail'></use></svg>
```
sreynen: This wasn't immediately clear to me, as I'd never seen this syntax before. I believe this is basically loading an SVG from elsewhere on the same page. That makes sense in the context of HTML, but for JSX we have the ability to load the SVG once but also have it in the DOM multiple times, and that seems worth doing to keep each component more self-contained. Does that make sense?

src/components/theme-giraffe/petition.js
```diff
@@ -0,0 +1,67 @@
...
+    <Message outOfDate={outOfDate} petition={p} isFwd={query.fwd} />
+    <InfoColumn>
+      <Card
+        heading={splitIntoSpansJsx(p.title)}
+        currentSignatures={p.total_signatures}
+        goalSignatures={p.signature_goal}
+      >
+        <p>To be delivered to {p.target.map(t => t.name).join(', ')}</p>
+
+        {p.featured_image_url && <Card.Media imageUrl={p.featured_image_url} />}
```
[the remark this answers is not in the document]
sreynen: I traced back through the code to the database to see where this is used and found it isn't used anywhere currently. So I think you can skip testing it for now, and I'll try to get some more info on history and future plans for this soon.

src/components/theme-giraffe/signature-add-form.js
```diff
@@ -0,0 +1,223 @@
...
+      // TODO: Needs design
+      <div>
+        <strong>Welcome back {user.given_name}!</strong>
+        <div>
+          (Not {user.given_name}? <a onClick={onUnrecognize}>Click here.</a>)
+        </div>
+      </div>
+    ) : (
+      // Anonymous
+      <div>
```
sreynen: The email and country inputs are overlapping in Firefox for me. I think maybe these container `<div>`s around the input blocks are the cause. Could we drop them?

src/giraffe-ui/petition/share.js
jparkrr: I originally did that, but ran into tons of issues. The svgs were misshapen, the wrong color, and didn't change on hover. You can see where I introduced the changes that match the comp here: https://github.com/MoveOnOrg/mop-frontend/pull/305/commits/9e0c153b8a6ddf6e17b790017803b76c25420615

I tried for a few hours to fix it with things like cleaning up the svgs and checking the css, but couldn't immediately find anything (plus I'm not an svg expert). So if you have any ideas or want to pair on this, let me know.
screen shot 2018-02-28 at 9 20 25 am

src/components/theme-giraffe/signature-add-form.js
jparkrr: I fixed this in https://github.com/MoveOnOrg/giraffe/pull/31, which wasn't merged before, but is now.

Even though it looks ok now, do you know a good way of removing the divs? I guess we could make all of the inputs into an array, but then you have to specify keys and it looks ugly IMO. Thoughts?

webpack.config.js
```diff
@@ -65,14 +66,15 @@ var config = {
                 : '/css/moui.css'),
       giraffeCssPath: (process.env.LOCAL_CSS
         ? 'http://localhost:3000/styles/main.css'
-        : 'https://static.moveon.org/giraffe/styles/main.css'),
+        : 'https://mop-static-stage.s3-us-west-1.amazonaws.com/giraffe/styles/main.css'),
```
schuyler1d: This needs to be configurable and at least one permutation should work with the production link :-)

Maybe just `giraffeCssPath: (process.env.LOCAL_CSS || 'https://mop-static-stage.s3-us-west-1.amazonaws.com/giraffe/styles/main.css'),` and then LOCAL_CSS can be a url

src/giraffe-ui/petition/share.js
jparkrr: Fixed it! Just had to pull the processed SVGs from the sprite. I tried to use this thing to do it automatically, but couldn't get the right settings https://github.com/svg/svgo

jparkrr:
Now dependent on https://github.com/MoveOnOrg/giraffe/pull/55

src/components/theme-giraffe/signature-add-form.js
sreynen: I'm not aware of another way to avoid container divs. I think some future version of React has a solution to that.

jparkrr:
Thanks for the reviews!

Issue #301, opened 2018-02-23
Signing doesn't yield r_hash or r_by links in e.g. email message


schuyler1d:
Steps to reproduce:
1. https://broom.moveon.org/sign/georgia-add-outkast-to
2. sign the petition. you get to https://petitions.moveon.org/thanks.html?petition_id=95983&name=georgia-add-outkast-to
which doesn't have an r_by.

Actually, I think this is because broom doesn't have `prodReady` on thanks, and so it moves to petitions.moveon.org instead of *broom.moveon.org`'s version of thanks page.

jparkrr:
So is the idea that, while `Thanks` isn't prodready, if you sign the petition at `https://broom.moveon.org/sign/georgia-add-outkast-to?r_id=something`, you should get forwarded to `https://petitions.moveon.org/thanks.html?petition_id=95983&name=georgia-add-outkast-to&r_id=something`?

jparkrr:
I guess easiest is if we are able to mark the Thanks page prodReady, as that is where all this logic was moved.

sreynen:
This was discussed in chat, but to close the loop here, no reason we can't mark Thanks as prodReady, as it is only ever reached via the sign page.

jparkrr:
Is this ok to leave closed now? Or do we want to verify that share tracking is working correctly?

PR #311, opened 2018-02-27
Mark Thanks route production ready

jparkrr:
Hopefully alleviates and/or fixes https://github.com/MoveOnOrg/mop-frontend/issues/301

Changing this means that submitting a signature will bring you to the reimplemented Thanks route instead of redirecting you to petitions.moveon.org/thanks.html.


jparkrr:
I verified locally that signing at least leads to an r_hash being set [here](https://github.com/MoveOnOrg/mop-frontend/blob/main/src/containers/thanks.js#L17)

PR #317, opened 2018-02-28
Update document title as route changes

### LINKED ISSUE #315: Update page <title> as route changes
For static pages, title comes from API. For everything else, maybe just hard-code them?

jparkrr:
fixes #315 

See https://github.com/gaearon/react-document-title for an explanation of the component I use. I think it's a pretty small dependency, and it fits pretty well with our container elements that aren't deeply nested.

The legacy system doesn't actually have that many titles, so I based most of these on the header of the page.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "osdi-client": "MoveOnOrg/osdi-client",
     "prop-types": "^15.5.8",
     "react": "^15.5.4",
+    "react-document-title": "^2.0.3",
     "react-dom": "^15.4.2",
     "react-redux": "^5.0.3",
     "react-responsive-embed": "^2.1.0",
--- a/src/components/theme-giraffe/wrapper.js
+++ b/src/components/theme-giraffe/wrapper.js
@@ -1,20 +1,23 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import DocumentTitle from 'react-document-title'
 import Nav from '../../containers/nav'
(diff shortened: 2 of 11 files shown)


jparkrr:
I'll leave this up for a while so people can look over the titles if they want.

jparkrr:
@schuyler1d right now I don't foresee too many problems, as long as we use `find()` in tests.

Here someone implemented a higher order component to do it: https://github.com/rrdelaney/react-page-title/blob/master/index.js, but it would still have to receive props from redux, so the end result would be similar, but with different syntax.

I think it's ok to go with the simplest solution until we see any problems, and then we can refactor. 

But if we really don't like wrappers, we can switch to [react-helmet](https://github.com/nfl/react-helmet), which doesn't actually require a wrapper and can be anywhere in the tree:
```jsx
<Parent>
    <Helmet>
        <title>My Title</title>
    </Helmet>

    <Child>
        <Helmet>
            <title>Nested Title</title>
        </Helmet>
    </Child>
</Parent>
```

schuyler1d:
ok, i'm :+1:  then -- I don't have any opinions on DocumentTitle vs. helmet -- choose the one that you (and/or @sreynen ) thinks best.

jparkrr:
@sreynen any thoughts?

sreynen:
This looks good to me.

jparkrr:
Thanks for the reviews!

PR #319, opened 2018-03-01
Fix nested routing

### LINKED ISSUE #318: Why don't static pages (or other path'd pages) load on stage site?
https://broom.moveon.org/2018/about.html
sets `window.baseAppPath = /2018/`  why doesn't that load the about content or load the react site?

jparkrr:
Turns out you have to leave of the leading slash in nested routes for it to respect the parent route, otherwise it just tries to match the entire path.

Seems weird to me, but I traced the code to find this:
https://github.com/ReactTraining/react-router/blob/07d6b50ab9a9b6d1940d67abe530044be7f7dc99/modules/matchRoutes.js#L126-L127

Fixes #318


jparkrr:
We didn't notice before because we had `/:organization/` routes for a lot of things that were actually loading.

Issue #327, opened 2018-03-05
Petition share links for redesign


schuyler1d:
On the thanks page, share links should be carried over from the legacy implementation.  Note that share links on that version are calculated partially from data gotten from the signature POST.

On the petition page itself, the new design includes four links:
 - [x] Facebook share - if there is a sharebandit url, upon clicking the link it should get `getSharebanditShareLink` the same way the thanks page does (ideally not before).  If there is no sharebandit link, it should be the standard facebook petition link, but with `source=n.fb.p` where `n` is for "non-signer" and `.p` is for the petition page.  Note that if the client 'knows' that the user is the creator or has signed the petition, then you can/should replace `n.` with either `c.` or `s.` appropriately
 - [x] Twitter share - This should use the lib `petitionShortCode` url, with mode=`n`.  Note that if the user is logged in or has guest credentials (e.g. akid, etc), then you can/should pass in the userId
 - [x] Email share - This will be the same thing as the thanks page except the link will be `source= n.mt.p` (unless it's meant to be copy-pasted, in which case n.cp.p)
 - [x] Copy Link share should be the same think as rawShareLink but with mode=`m` for a non-signer (keep the other two, if they are known creator/signers, but don't hit an additional api to find out).

See [this PR](https://github.com/MoveOnOrg/mop/pull/235) and the full shortcodes file for short-code related questions. 

Also update docs/EXPLANATION-share-links.md

jparkrr:
For the sign page we might want a slight rephrasing of the mail message, it currently says

> That's why I **signed** a petition to President Donald Trump, The United States House of Representatives, and The United States Senate, which says:

jparkrr:
For email share on the thanks page we have something like `source=s.em.mt`. Did you drop "em" on purpose above for the petition page? If I kept it, it would be like `n.em.mt.p`

schuyler1d:
that was a mistake.  do `n.em.mt.p`

jparkrr:
Thanks! Last question I think: should we also `recordShareClick()` for the sign page?

PR #330, opened 2018-03-06
Refactor social sharing

jparkrr:
Will enable https://github.com/MoveOnOrg/mop-frontend/issues/327

I extract the sharing logic from the Thanks container, and create [higher-order components](https://reactjs.org/docs/higher-order-components.html) that wrap the actual share buttons, so this sharing behavior can be used in multiple places.

I think it also makes it clearer how things are supposed to be working.


jparkrr:
Ready for review! It would be cool to add tests. I'll come back to that if I get a chance.

src/components/theme-legacy/twitter-button.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import { withTwitter } from '../../containers/hoc-twitter'
+
+const TwitterButton = ({ onClick: shareTwitter }) => (
+  <button
+    id='twitter-button'
+    className='xl300 background-moveon-bright-red'
+    onClick={shareTwitter}
```
schuyler1d: Can you explain how this `onClick` functions wrt the onClick that is in `withTwitter`?  even if it all works out here, I feel like it might be a source of bugs if we don't make it clear who/what is happening on the click event

jparkrr: I rename onClick to shareTwitter here:
https://github.com/MoveOnOrg/mop-frontend/blob/9ba07ebaaca1bf8473a3f273230e56eea83dd308/src/components/theme-legacy/twitter-button.js#L6

Is that what you mean? I should probably standardize fully on onClick though

schuyler1d: ah, ok, I think ironically the rename makes it more confusing.  maybe just do `({ onClick })` and then it's more obvious it comes from someplace else

jparkrr:
Thanks for the review @schuyler1d!

Are you suggesting that you think all of the functions that construct the sharing links/messages should be pulled out somewhere else?

If so, I think that would be pretty cool if there was more shared logic between those functions, but they're currently pretty separate.

I'd say leave it in the component for now and see how it goes once there are some updates that need to be made to sharing. Probably a lot of sharing logic is service-specific. Not sure though.

Or let me know if you had something else in mind.

schuyler1d:
That is what I meant.  I understand that the logic ends up being independent, but it seemed like it would be nice to open 'one file' and see all the ways share links are constructed and what contexts they need. -- e.g. for future folks (including me) that are trying to wrap their heads around all the docs/EXPLANATION--share-links.md things.

If you think it's not worth it, then I'll defer to you.

jparkrr:
Just my opinion, but having semi-related logic in one file doesn't make as much sense as it used to in the world of components. I'm still all for extracting similar logic though.

PR #332, opened 2018-03-06
get signatures by list id rather than slug

### LINKED ISSUE #331: Query comments by petition.list_id when available
The database query on the backend is much more efficient when we query comments by list id, so we should try to do that.

schuyler1d:
fixes #331

My guess is that this is not the most elegant code for react/es6 style.  I'd love feedback on how this would be better written in a regular style.

### DIFF
--- a/src/components/theme-legacy/petition.js
+++ b/src/components/theme-legacy/petition.js
@@ -8,11 +8,14 @@ import SignatureList from '../../containers/signature-list'
 import PetitionFlagForm from '../../containers/petition-flag-form'
 import { Link } from 'react-router'
 
-const Petition = ({ petition: p, query, petitionBy, outOfDate }) => {
-  const petitionListIds = p.identifiers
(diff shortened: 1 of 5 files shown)


src/components/theme-legacy/petition.js
```diff
@@ -8,6 +8,13 @@ import SignatureList from '../../containers/signature-list'
 import PetitionFlagForm from '../../containers/petition-flag-form'
 import { Link } from 'react-router'
 
+function getPetitionListId(petition) {
```
jparkrr: Maybe move this function into `src/containers/signature-list.js` and pass `p.identifiers` into signature list? It would be cool to have a comment about what this function does also.

schuyler1d: Ok, moved it to actions, which I think hopefully achieves the same result.  note that I also updated theme-giraffe/petition.js too

PR #335, opened 2018-03-07
some general doc cleanup

sreynen:
This is an attempt to make the project a little more approachable to new contributors.


docs/HOWTO--server-backend.md
```diff
@@ -0,0 +1,15 @@
+# How to Develop with a Server Backend
+
+When running locally, you should set the following environment variables:
+
+```
+export API_URI="https://petitions.example.com"
+export BASE_APP_PATH="/Users/yourusername/Sites/mop-frontend/local/"
```
sandramchung: This should be generic-ized to something like `export BASE_APP_PATH="/path/to/your/local/mop-frontend/local/"`

docs/HOWTO--server-backend.md
```diff
@@ -0,0 +1,15 @@
...
+export STATIC_ROOT="../../build/"
+```
+
+`API_URI` can either point at a hosted version of the API (as in the example), or a local instance of mop (something like `http://0.0.0.0:8000`).
```
sandramchung: They don't have access to mop yet. Can they use broom.moveon.org? Do we note that here or mention it to non-MO devs as needed?

shakalee14: same question

sreynen: I was thinking mention as needed.

sreynen: Just realized we can just use petitions.moveon.org as the example now that the API is live there.

docs/HOWTO--server-backend.md
schuyler1d: is that actually necessary at all?  I think BASE_APP_PATH's default is your local env, which should be fine, right?

docs/HOWTO--server-backend.md
```diff
@@ -0,0 +1,15 @@
+# How to Develop with a Server Backend
+
+When running locally, you should set the following environment variables:
+
+```
+export API_URI="https://petitions.moveon.org"
+export BASE_APP_PATH="/path/to/your/local/mop-frontend/local/"
```
schuyler1d: Is BASE_APP_PATH necessary?  if folks are running with just `npm run dev` I think it should set it correctly by default.

Issue #309, opened 2018-02-26
Apply redesign to sign page


sreynen:
- [x] Overall page style (#305)
- [x] Design updates for the signature list (#307)
- [x] Add the mobile sign button and modal form (#322)
- [x] Working share buttons (#336)
- [x] "Thank you" page style (#333)
- [x] replace modal with inline form (#384 )
- [x]  petition messages (#383)
- [x] Thanks redirect message (#383)

jparkrr:
Related issue for mobile signing: https://github.com/MoveOnOrg/giraffe/issues/76

jparkrr:
The basic work has been done. We can file new issues if we find anything else to do.

PR #333, opened 2018-03-06
Giraffe thanks page

jparkrr:
Depends on https://github.com/MoveOnOrg/mop-frontend/pull/330

Restyles the thanks page for giraffe
http://localhost:8080/thanks.html?petition_id=95983

tracking issue #309

PR #336, opened 2018-03-07
Giraffe sign page social sharing

### LINKED ISSUE #327: Petition share links for redesign

jparkrr:
Depends on #330

Makes the social sharing links on the sign page work. See #327 for the explanation of the source prefixes and suffixes.

closes #327 
tracking issue #309


src/components/theme-giraffe/petition-share.js
```diff
@@ -0,0 +1,55 @@
...
+))
+
+const Facebook = withFacebook(MoShare.Facebook)
+
+const Twitter = withTwitter(MoShare.Twitter)
+
+const CopyLink = withShareLink(({ rawLink, ...rest }) => (
+  <MoShare.CopyLink
+    {...rest}
+    onClick={() => prompt('Here is your share link:', rawLink)}
```
jparkrr: I asked a question about this here: https://github.com/MoveOnOrg/giraffe/issues/79

sreynen:
Marking this as blocked on the UX issue. Also has the same issue with short links I flagged in #333. Otherwise looks good to me.

sreynen:
Looks like this is no longer blocked.

jparkrr:
Added copy link here too. Should be good to go once https://github.com/MoveOnOrg/giraffe/pull/111 and https://github.com/MoveOnOrg/mop-frontend/pull/333 are merged

src/giraffe-ui/petition/share.js
```diff
@@ -7,39 +7,17 @@ import TwitterSvg from '../svgs/twitter.svg'
...
-        </Link>
-        <Link onClick={tweet} className={elementCn}>
-          <TwitterSvg />
-          {hasLabels && 'Twitter'}
-        </Link>
-        <Link onClick={link} className={elementCn}>
-          <LinkSvg />
-          {hasLabels && 'Copy Link'}
-        </Link>
+      <div className={`${className}__share${hasLabels ? '__links' : '-icons'}`}>
```
schuyler1d: this is another place where it might be useful to revisit the classname-choices as a library for the new theme.

PR #341, opened 2018-03-07
Fix a reducer state mutation

jparkrr:
Found by our  nice devtool merged in #312

### DIFF
--- a/src/reducers/index.js
+++ b/src/reducers/index.js
@@ -104,7 +104,7 @@ function petitionReducer(state = initialPetitionState, action) {
         ...updateData
       }
     case petitionActionTypes.FETCH_PETITION_SIGNATURES_SUCCESS:
-      petition.total_signatures = signatures.count
+      petition = { ...petition, total_signatures: signatures.count }
       return {
         ...state,
         petitionSignatures: {


jparkrr:
@schuyler1d yeah it's here
https://github.com/MoveOnOrg/mop-frontend/blob/af6171bcc056d13b1414cd0ce0588280e7f42640/src/reducers/index.js#L106-L107

I think it's nice to make a copy as soon as possible, I don't think it should be slow or anything, but I could be misunderstanding also

schuyler1d:
So my understanding is that a bunch of react is about not updating objects unless something has changed.  So if we put it where this PR does, basically every time anything related to a petition is called React's logic will try to push down/update DOM state for anything tied to a petition, but will skip that if the petition isn't changed.

For the modification site lower down, I think maybe we should track total_count in a separate state object, and then update that directly.  I was going to say that maybe we don't need this update to total sigs at all, because we already get it from the petition load page, but we might cache the load page longer than the signature data, and it would be nice to be able to keep petition signature count 'fast updatable'

jparkrr:
@schuyler1d ah yeah, you're totally right. I wasn't remembering correctly how react-redux works. I thought it somehow did a deep comparison to know about updates, but it's actually a shallow comparison. So it is indeed better to return the original object if possible. https://redux.js.org/faq/react-redux 

I updated it to only make a copy of the petition when we fetch signatures. Do you think I should pull out total_signatures now, or maybe at some later time? I think it's unlikely we'll run into any performance problems due to rerendering soon.

sreynen:
Looking at the diff on this, it seems like it now only includes the change to address @schuyler1d's concern, but is missing code that was there before?

jparkrr:
@sreynen right, so the patten now is this line will stay the same and `petition` will contain a reference to the petition in the state
https://github.com/MoveOnOrg/mop-frontend/blob/3a9a268706f44dbd69988fbd91e9c825c1898c5e/src/reducers/index.js#L69-L70

but if we need to modify that petition, we will make a copy at that point
https://github.com/MoveOnOrg/mop-frontend/blob/3a9a268706f44dbd69988fbd91e9c825c1898c5e/src/reducers/index.js#L107
which will alleviate the error because we are no longer modifying state directly

PR #344, opened 2018-03-08
Rewrite history match location in include baseAppPath

### LINKED ISSUE #343: Thanks page still redirecting on broom when using baseAppPath
signing a petition on broom should stay on broom/react when moving to the thanks page, but even though there is `prodReady` marked on the route, when I sign a petition (e.g. the georgia outkast one), I get directed on broom to petitions.moveon.org

jparkrr:
This uses our overwritten <Link> routes and router.push() matcher to rewrites match locations to include baseAppPath (like /2018/). This means you should stay within the baseAppPath you originally accessed.

If a route is not found or not prodReady, we redirect using the original path  like petitions.moveon.org/thanks.html (without /2018/)

Fixes #343

### DIFF
--- a/src/routes.js
+++ b/src/routes.js
@@ -40,7 +40,7 @@ const updateHistoryObject = (historyObj, routes) => {
   // eslint-disable-next-line no-param-reassign
   historyObj.push = (path, state) => {
     let matchPath = path
-    if (baseAppPath !== '/') {
+    if (baseAppPath !== '/' && path.substring(0, 4) !== 'http') {
       matchPath = baseAppPath + (path[0] === '/' ? path.substring(1) : path)
     }
     match(
--- a/src/routes.js
+++ b/src/routes.js
@@ -19,14 +19,19 @@ import Login from './containers/login'
 import Static from './containers/static'
 
 
-const baseAppPath = process.env.BASE_APP_PATH || '/'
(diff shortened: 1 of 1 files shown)


src/routes.js
```diff
@@ -19,28 +19,37 @@ import Login from './containers/login'
...
   // All <Link>s and appLocation.push calls in the codebase should NOT be relative
   // -- i.e. they should be absolute paths like /thanks.html
 
   const PROD_URL = 'https://petitions.moveon.org'
   const origPush = historyObj.push
   // eslint-disable-next-line no-param-reassign
   historyObj.push = (path, state) => {
+    let matchPath = path
+    if (baseAppPath !== '/') {
+      matchPath = baseAppPath + (path[0] === '/' ? path.substring(1) : path)
```
schuyler1d: can path ever be `http://...` -- even if React doesn't think so, I think we use it that way.  e.g. for https://github.com/MoveOnOrg/mop-frontend/pull/344/files#diff-3ca5e313cdf02368ce63986a36207c94R57

jparkrr: yeah, you're right. I've unfortunately been going back and forth on using <Link/>s for external links.

schuyler1d: I'm pretty sure we have at least some use-cases for it (forget what they are -- at least to make `nav` components simpler, I think).  so worth just guarding against

PR #345, opened 2018-03-08
Use browser routing by default in dev

jparkrr:
It's good because it matches production more closely.

Will default to browser routing in dev and prod.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -15,6 +15,9 @@ var config = {
   },
   devServer: {
     host: "0.0.0.0",
+    historyApiFallback: {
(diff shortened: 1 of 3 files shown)


webpack.config.js
```diff
@@ -15,6 +15,9 @@ var config = {
   },
   devServer: {
     host: "0.0.0.0",
+    historyApiFallback: {
+      disableDotRule: true
```
jparkrr: In case you're interested in what this means: by default it looks on the filesystem for paths with a `.` in them, so our .html files wouldn't be passed to react without this.

PR #346, opened 2018-03-09
Add /pac/ route, header, and footer

### LINKED ISSUE #323: Make /pac/ load PAC home page
Should be same as normal home page, but with pac=1 in top petitions API request.

### LINKED ISSUE #324: Add PAC disclaimer to PAC petitions
E.g. /sign/its-time-to-reverse-decades should show

"This tool was licensed to and paid for by MoveOn.org Political Action. Not authorized by any candidate or candidate's committee. Terms of Service."

like https://pac.petitions.moveon.org/sign/its-time-to-reverse-decades does.

### LINKED ISSUE #325: Show PAC "Political Action Edition" logo subtext on PAC petitions and /pac/ home page
See https://pac.petitions.moveon.org/sign/its-time-to-reverse-decades for example.

(Sky) Note: this is only a difference in the legacy theme -- the rebrand has no difference.

jparkrr:
Some of this logic existed already, but was implemented by treating 'pac' as an organization. This seemed to cause a number of problems, so I removed that and created a component `PacHome` that wraps `Home` and sets the `isPac` prop.

I removed some logic in TopPetitions that stopped it from firing the `loadTopPetitions` action if we already had top petitions. I think this is safe to remove because we already cache the top petitions in the reducer.

closes #323 

But we also have to provide `entity=pac` to the header and footer, which we have to do in an unrelated way (I think). We create a Wrapper container, which sets `entity` on the Header and Footer in two ways
- Based on the `location` prop provided by the router, checking if we are on the pac homepage.
- Based on the entity of the petition loaded from the state, if the route has a `petition_slug` param.
closes #325 
closes #324 

Let me know if you can think of a better way of doing any of this.


src/containers/home.js
```diff
@@ -25,7 +24,7 @@ const Home = ({ params, nav }) => {
 
       <div className='row front-content'>
         <TopPetitions
-          pac={isPac ? 1 : 0}
+          pac={isPac || Config.ENTITY === 'pac'}
```
schuyler1d: I think we can get rid of Config.ENTITY now that we are collapsing it to one site.  Originally, it was (feared?) that we'd have to build two different react apps.

src/components/theme-legacy/billboard.js
```diff
@@ -5,7 +5,7 @@ import { Link } from 'react-router'
 const BillBoard = () => (
   <div className='container'>
     <div className='row'>
-      <div className='span12 billboard-background bump-bottom-1 clearfix'>
+      <div className='span12 billboard-background bump-bottom-1 clearfix' style={{ float: 'left' }}>
```
jparkrr: I had to add this because, when the PAC header is showing, it was overlapping with `billboard-background`, which was set to not float. I think it should be safe since span12s normally float left, but I don't know why float:none was added. It seems to work everywhere in my testing.

src/containers/top-petitions.js
```diff
@@ -36,19 +34,20 @@ TopPetitions.propTypes = {
 
 const mapStateToProps = (store, ownProps) => {
   const { pac, megapartner } = ownProps
-  const topPetitionsState = (store.petitionStore.topPetitions || {})
-  // TopPetitionsKey must not just be truthily equal but exact
```
jparkrr: Not sure what this comment means

jparkrr:
Thanks for the review!

PR #352, opened 2018-03-14
Show only 50 states in petition target dropdown

jparkrr:
Gives the bool param `onlyStates` to StateSelect.

So you can only target the elected officials of the 50 states, matching legacy.

We also use StateSelect for search and signing, to which we don't pass `onlyStates`, so it has other regions, matching legacy.

### DIFF
--- a/src/components/theme-legacy/form/state-select.js
+++ b/src/components/theme-legacy/form/state-select.js
@@ -1,6 +1,6 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import { states, armedForcesRegions } from '../../../lib/state-abbrev'
+import { getRegions, armedForcesRegions } from '../../../lib'
 
 const StateSelect = ({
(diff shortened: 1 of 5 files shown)


jparkrr:
Good catch @sreynen. Fixed.

schuyler1d:
Was this a policy decision? It seems odd that we would eg block petitions created in/around puerto rico, and thus possibly buy extension the others.

jparkrr:
@schuyler1d I was hoping someone would say that 👍 

But this does match legacy, and it would require some extra work on the backend to get legislators from Puerto Rico https://petitions.moveon.org/target_search.html?group=state&state=PR

Looks like they do have a Senate and House though, so maybe I could just leave off the legislator autocomplete if no results are returned? https://en.wikipedia.org/wiki/Legislative_Assembly_of_Puerto_Rico

PR #353, opened 2018-03-14
add nyc and test:coverage, remove (unused?) test1

sreynen:
This gives us a command to check test coverage, as a step toward improving coverage. It also removes the test1 script. Looking at the commit history, I'm not clear why test1 was originally added, so using this PR to at least clarify that, if not remove it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "repository": "https://github.com/MoveOnOrg/mop-frontend",
   "directories": {},
   "scripts": {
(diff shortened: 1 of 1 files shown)


jparkrr:
To remove `test1` but retain the functionality, we could create a file test/mocha.opts with just
```
--recursive
```

and then you should be able to run one file using `npm test -- path/to/file.js`, which appends the argument

schuyler1d:
Thats fine for me. Can we document that in readme or contributing -- wherever appropriate

jparkrr:
or actually we don't even need mocha.opts, we could just put --recursive on the test script (because looking into the /test dir is the default

jparkrr:
Pushed my suggested changes

PR #354, opened 2018-03-14
WIP: Petition creation part 1: target autocomplete, add to redux store

### LINKED ISSUE #202: Add target search to petition create form

- sample api url: https://broom.moveon.org/api/legacy/target_search.html?search=char&group=senatemem recreate autocomplete interface

jparkrr:
One difference from legacy is it doesn't remove targets you've added from the UI when you uncheck them, but I can add a special case for this if we want. I was trying to make these added targets behave exactly like the preloaded, unchecked targets, which are stored in the same state variable.

closes #202


jparkrr:
OK, I don't think we should merge until https://github.com/MoveOnOrg/mop/issues/263 is fixed. Let me know if you disagree

schuyler1d:
@jparkrr  Is everything that needs to be merged, merged in this branch?  If so, let's close.  If not, let's resolve conflicts.

schuyler1d:
ah, didn't realize this was going to the login branch.  if this is still WIP, let's mark it that way

jparkrr:
This work is in #531

PR #355, opened 2018-03-14
Validate create petition, allow target autocomplete, and add to redux store

### LINKED ISSUE #202: Add target search to petition create form

### LINKED ISSUE #203: Validate and submit petition create form step 1

jparkrr:
depends on #352 and #354 

Sorry about the large PR
Ready for review I think. Especially try out the petition target autocomplete.

One difference from legacy is it doesn't remove targets you've added from the UI when you uncheck them, but I can add a special case for this if we want. I was trying to make these added targets behave exactly like the preloaded, unchecked targets, which are stored in the same state variable.

I did it this way because it should be fairly easy to just submit the entire redux stored petition object to the backend. (future PR)

Open to any ideas or feedback.

fixes #202


package.json
```diff
@@ -19,6 +19,7 @@
   "license": "MIT",
   "dependencies": {
     "classnames": "^2.2.5",
+    "downshift": "^1.30.0",
```
jparkrr: one cool thing about downshift (autocomplete component) is it's accessible by default https://github.com/paypal/downshift/blob/07b0d953f678b7536821794a3614dce4e092018a/src/utils.js#L181

src/components/theme-legacy/form/target-select/legislator-autocomplete.js
```diff
@@ -0,0 +1,88 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import Downshift from 'downshift'
+
+const container = {
```
jparkrr: on legacy, these styles come from jquery-ui... I just threw the relevant ones in this file because I see no reason they should be reusable

src/containers/create-petition-target.js
```diff
@@ -0,0 +1,203 @@
...
+              name: 'target',
+              value: this.getAllCheckedTargets()
+            }
+          })
+        )
+    }
+  }
+
+  getAllCheckedTargets() {
+    const groups = [this.state.national, this.state.geoStateSelected]
```
jparkrr: Oops, need to add custom targets here

src/containers/create-petition-target.js
```diff
@@ -0,0 +1,203 @@
...
+    if (!nationalOpen) return null
+    return (
+      <NationalTargetSelect
+        selected={national}
+        onSelect={this.onSelect('national')}
+      />
+    )
+  }
+
+  renderGeoState() {
```
jparkrr: Do you like the name `geoState`? I use it to differentiate a U.S. State from component state.

src/components/theme-legacy/form/target-select/legislator-autocomplete.js
```diff
@@ -0,0 +1,89 @@
...
+  marginTop: '-11px',
+  padding: '2px',
+  borderRadius: '4px',
+  position: 'absolute',
+  border: '1px solid #aaaaaa',
+  background: '#fff',
+  backgroundImage:
+    'url(http://ajax.googleapis.com/ajax/libs/jqueryui/1.8.9/themes/base/images/ui-bg_flat_75_ffffff_40x100.png) 50% 50% repeat-x',
```
sreynen: This should probably use https:// URL, which seems to work.

src/components/theme-legacy/form/target-select/legislator-autocomplete.js
```diff
@@ -0,0 +1,89 @@
...
+  color: '#222222',
+  fontFamily: 'Verdana,Arial,sans-serif',
+  fontSize: '1.1em',
+  cursor: 'default'
+}
+
+const hover = {
+  border: '1px solid #999999',
+  background:
+    '#dadada url(http://ajax.googleapis.com/ajax/libs/jqueryui/1.8.9/themes/base/images/ui-bg_glass_75_dadada_1x400.png) 50% 50% repeat-x',
```
sreynen: Same here, let's use https:

src/containers/create-petition-target.js
sreynen: 👍

PR #361, opened 2018-03-16
Make sure ThanksNextPetition interval is set on focus

### LINKED ISSUE #359: "We've found another petition which may interest you" countdown gets stuck
For me it currently says "We'll forward you there in 11 seconds" and isn't changing.

I had to remind myself how this is supposed to work. Here's the comment explaining it: https://github.com/MoveOnOrg/mop-frontend/blob/main/src/containers/thanks-next-petition.js#L8

jparkrr:
If nextPetition (from a top-petitions request) was not loaded when the window focused after sharing (probably because of a slow request), the countdown would not start.

This changes it to start the background interval regardless of the top-petitions request. However, if there is some error with that request, the message still won't be shown (guarded in the render method) and they won't be redirected when the interval finishes (guarded in `redirectToNext()`)

Fixes #359

### DIFF
--- a/src/actions/petitionActions.js
+++ b/src/actions/petitionActions.js
@@ -399,5 +399,6 @@ export const actions = {
   registerSignatureAndThanks,
   recordShareClick,
   loadPetitionSignatures,
-  getSharebanditShareLink
+  getSharebanditShareLink,
+  loadTopPetitions
 }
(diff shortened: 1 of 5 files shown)


jparkrr:
@sreynen I don't remember what we decided in chat. Do you want to let me know where we landed or take over here?

jparkrr:
@sreynen I went ahead and moved loadTopPetitions into the thanks container

PR #362, opened 2018-03-16
Make thanks.html redirect message appear at top of screen

### LINKED ISSUE #360: "We've found another petition which may interest you" banner needs to go above header
I suspect this is going to be complicated to do, but seems important to make this notice more prominent since we're redirecting the page.

jparkrr:
Use [react-portal](https://github.com/tajo/react-portal) to render ThanksNextPetition in a div included in the Wrapper

Fixes #360

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "prop-types": "^15.5.8",
     "react": "^15.5.4",
     "react-dom": "^15.4.2",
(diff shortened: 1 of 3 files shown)


jparkrr:
I moved the portal element into theme-legacy/wrapper.js :)

PR #379, opened 2018-03-20
Track on history push and pop, pass page as pathname

### LINKED ISSUE #369: trackPage just tracking canonical url
When I go to https://petitions.moveon.org/2018/sign/stop-censoring-backyard

I get the following in GA debugger (snipped):
```
VM298 analytics_debug.js:10 language         (&ul)   en-us
VM298 analytics_debug.js:10 location         (&dl)   https://petitions.moveon.org/2018/sign/stop-censoring-backyard
VM298 analytics_debug.js:10 page             (&dp)   /sign/stop-censoring-backyard
VM298 analytics_debug.js:10 screenColors     (&sd)   24-bit
```
Note that location is the full/real one, but page is the canonical url.
Let's get it to work and send page correctly here:
https://github.com/MoveOnOrg/mop-frontend/blob/main/src/actions/sessionActions.js#L91
(in both branches)

jparkrr:
fixes #369 so we should have the full url in `page`

We want to track regardless of the action, because pushes will fire when they navigate to a new page, and pops (only) will fire when they go back

### DIFF
--- a/src/actions/sessionActions.js
+++ b/src/actions/sessionActions.js
@@ -82,21 +82,22 @@ export const loadSession = ({ location }) => {
 
 export const trackPage = () => {
   // This will track both when they navigate to a new page OR use the back button
-    if (window.analytics || window.gtag) {
-      const href = String(document.location.href)
-        .replace(/([?&])source=/, (full, op) => `${op}utm_source=`)
-      if (window.analytics) {
-        // Segment.com tracking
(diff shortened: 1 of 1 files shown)


jparkrr:
@schuyler1d super weird, I can't reproduce that in chrome or firefox...

jparkrr:
yup, for posterity: turned out to be a difference between hashrouter and pagerouter https://github.com/ReactTraining/history/issues/427

PR #380, opened 2018-03-20
Move loadSession to wrapper so it only fires once

### LINKED ISSUE #348: on /2018/ urls session.json is called for each page visit (should only be initial load)
Go to /2018/ click on a petition.  Note that two session.json calls are made.  I believe we should only make the initial one.

jparkrr:
First I tried refactoring routes() to a class, but it didn't work out since the current location wasn't available outside <Router> 🚯 

This should still be fairly discoverable, with the comment in loadSession

fixes #348

### DIFF
--- a/src/actions/sessionActions.js
+++ b/src/actions/sessionActions.js
@@ -54,8 +54,8 @@ function callSessionApi(tokens) {
   }
 }
 
-export const loadSession = ({ location }) => {
-  // Called straigt from top route onEnter, so it's done only once on first-load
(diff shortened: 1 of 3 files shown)


jparkrr:
@schuyler1d yes, entities can be set from the url, or the petition currently loaded here https://github.com/MoveOnOrg/mop-frontend/blob/a8d6516bd3cc5fa22548cd986c5408624a9a3182/src/containers/wrapper.js#L42-L50

PR #385, opened 2018-03-26
Fixes for the giraffe sign form

jparkrr:
- Don't include moui in giraffe in dev
- Correct design for checkboxes in giraffe sign form
- Make validation text red
- Giraffe styles don't need value from state anymore
- Use helper methods for setting thirdparty_optin in state


jparkrr:
We might want to do more on the design for validation. Would be nice to not stray too far from legacy though. cc @sandramchung ? 

Also we may want to re-test things like collect_volunteers or thirdparty_optin before launching. I'm not sure if that's part of QA?

PR #387, opened 2018-03-27
Pass thirdPartyOptin correctly to sign form, handle toggling correctly

jparkrr:
Previously, we weren't passing the value in the state to be set as the checkbox's default value.

Additionally, toggling of the checkbox wasn't updating the state correctly.

### DIFF
--- a/src/components/theme-giraffe/signature-add-form.js
+++ b/src/components/theme-giraffe/signature-add-form.js
@@ -144,7 +144,7 @@ const SignatureAddForm = ({
         name='thirdparty_optin'
         type='checkbox'
         label={`Receive campaign updates from ${creator.organization || 'this organization'}`}
-        onChange={updateStateFromValue('thirdparty_optin', true)}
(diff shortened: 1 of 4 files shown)


jparkrr:
It also seems like deselecting the checkbox isn't updating the value in the state correctly. Will push a fix soon.

jparkrr:
all fixed

src/components/theme-giraffe/signature-add-form.js
```diff
@@ -143,7 +144,8 @@ const SignatureAddForm = ({
         name='thirdparty_optin'
         type='checkbox'
         label={`Receive campaign updates from ${creator.organization || 'this organization'}`}
-        onChange={updateStateFromValue('thirdparty_optin')}
+        onChange={updateStateFromValue('thirdparty_optin', true)}
```
schuyler1d: i don't know if our ES/compiling version accepts named args yet, but we do at least something like `/*isCheckbox=*/` here to be explicit about what is 'true' (ditto for the other place(.

PR #389, opened 2018-03-28
Refactor to prepare for giraffe home

jparkrr:
This PR should have no effect on how anything looks, either in giraffe (sign page) or legacy, just some refactoring to prepare for the giraffe petition home page.

- Extract reusable Card component from PetitionCard
- Extract stateless home component
- Load recent victories from an object and refactor
- Rename TopPetition's PetitionPreview to Blurb and extract stateless component


jparkrr:
If you want to test giraffe sign page locally, you'll need this checked out: https://github.com/MoveOnOrg/giraffe/pull/135

I don't want to merge it until this PR is approved because it will break the live giraffe petition sign page.

PR #391, opened 2018-03-28
Add email button tracking on thanks page

jparkrr:
We noticed this was missing

### DIFF
--- a/src/containers/thanks.js
+++ b/src/containers/thanks.js
@@ -111,6 +111,7 @@ class Thanks extends React.Component {
         petition={this.props.petition}
         prefix={this.state.pre}
         trackingParams={this.trackingParamsString}
+        recordShare={this.recordShare('email', `${this.state.pre}.em.mt`)}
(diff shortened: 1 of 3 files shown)


src/containers/thanks.js
```diff
@@ -111,6 +111,7 @@ class Thanks extends React.Component {
         petition={this.props.petition}
         prefix={this.state.pre}
         trackingParams={this.trackingParamsString}
+        recordShare={this.recordShare('email', `${this.state.pre}.em.mt`)}
```
sreynen: Doesn't this need to be `onClick` to match the prop name?

jparkrr: Actually, since it passes through `withShareMessage()`, this is correct https://github.com/MoveOnOrg/mop-frontend/blob/bdd81f9df6c7b46d117c22d45ea7219ac735c92b/src/containers/hoc-share-message.js#L93-L98

PR #396, opened 2018-03-29
Fix logic to hide sign form address when user is signing petition

### LINKED ISSUE #395: Recognized user is prompted for address on non-local petition, should not be
Compare https://petitions.moveon.org/2018/sign/tell-facebook-stop-enabling?ak_proof=1&akid=202494.11159533.qi-VxT&rd=1&source=mop&t=13 vs. https://petitions.moveon.org/sign/tell-facebook-stop-enabling?ak_proof=1&akid=202494.11159533.qi-VxT&rd=1&source=mop&t=13

Not entirely sure this is a frontend bug, but suspect it is.

jparkrr:
fixes #395

### DIFF
--- a/src/containers/signature-add-form.js
+++ b/src/containers/signature-add-form.js
@@ -250,16 +250,27 @@ SignatureAddForm.propTypes = {
   setRef: PropTypes.func
 }
 
+function showAddressFields(user, petition) {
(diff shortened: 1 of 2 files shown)


jparkrr:
need to fix the tests

PR #398, opened 2018-03-30
Hide floating button when sign form is visible

jparkrr:
Uses [react-waypoint](https://github.com/brigade/react-waypoint) to hide the floating sign button when the form is visible. It transitions onto screen when you scroll below the form.

See https://github.com/MoveOnOrg/giraffe/issues/129 for discussion.

Requires https://github.com/MoveOnOrg/giraffe/pull/155


src/components/theme-giraffe/signature-add-form.js
```diff
@@ -22,15 +22,18 @@ const SignatureAddForm = ({
   onUnrecognize,
   updateStateFromValue,
   validationError,
-  setRef
+  setRef,
+  innerRef
 }) => (
-  <form onSubmit={submit} className='sign-form'>
+  <form ref={innerRef} onSubmit={submit} className='sign-form'>
```
jparkrr: need to have innerRef for react-waypoints to work

PR #403, opened 2018-04-03
Make a webpack chunk for every route

jparkrr:
I had planned to only split some things out into chunks, but [react-loadable](https://github.com/jamiebuilds/react-loadable) was so easy, I thought it made more sense to use it for everything. So now we have 10 chunks.

We also split out polyfills into another file, which should be conditionally included in the html.

As @schuyler1d suggested, I kept the sign page in the main bundle.

I created a loading component that is displayed when a chunk is loading.

I switched to using chunkHash instead of just hash, so the chunks can remain cached if they don't change.

takes the prod main bundle size from 273k -> 177k
gzipped 67k -> 45k

Requires https://github.com/MoveOnOrg/mop/pull/342 on the backend

TODO:
- [x] Move sign things to main bundle
- [x] Make it work in ie 11
- [x] Try to break out polyfills based on https://webpack.js.org/guides/shimming/#loading-polyfills


jparkrr:
It's a little bit difficult to remove thanks-shim because Thanks container is written in such a way that assumes `petition` is loaded (because it's loaded in thanks-shim). Totally possible to fix, but I don't want to spend any more time on this right now.

Let me know if you have any thoughts on the savings.

jparkrr:
@schuyler1d No not really, 80% of chunk-0 is the containers and theme components to sign the form, the rest is react-portal.  This is the biggest chunk at 50k, but I'm fine with including it in the bundle for now.

jparkrr:
I should also rerun all these numbers for giraffe, since that's the longer term thing. There are a couple different modules for that on the sign-page.

jparkrr:
We would need to load polyfills before the bundle, doesn't look too bad, some instructions here https://webpack.js.org/guides/shimming/#loading-polyfills

React 16 wouldn't make too much of a difference. Webpack 4 bundles a little bit differently, but it shouldn't make a difference here (it might give us fewer chunks).

jparkrr:
I just tested and this currently breaks in ie 11 also, probably because the import() needs polyfills earlier, which is also mentioned in https://webpack.js.org/guides/shimming/#loading-polyfills

jparkrr:
If you want to read about what webpack 4 will do check out https://gist.github.com/sokra/1522d586b8e5c0f5072d7565c2bee693, but only if you have a lot of time, because it's kinda involved :)

jparkrr:
Ok, everything is all updated and I did a good amount of testing (including ie 11 pointed at local dev)

See #429 for the bundle analyzer plugin changes.

src/apps/main.js
```diff
@@ -2,17 +2,22 @@ import React from 'react'
...
+    document.getElementById('root')
+  )
+}
 
-ReactDOM.render(
-  <Provider store={store} children={routes(store)} />,
-  document.getElementById('root')
-)
+const loadedStates = ['complete', 'loaded', 'interactive']
+if (loadedStates.includes(document.readyState) && document.body) {
```
schuyler1d: rather than add a polyfill for IE for Array.includes, maybe we can just do 3 consecutive OR statements or 
```
const loadedStates = {complete:1,loaded:1,interactive:1}
if (document.readyState in loadedStates)...
```

jparkrr: Thanks. fixed

PR #406, opened 2018-04-04
 Register and login

### LINKED ISSUE #148: Make login form POST login credentials to API, recieve auth cookie

jparkrr:
fixes #148 

I added a proxy to port 8000 in webpack dev server, so you can use a running local mop with `export API_URI="/api/v1"`

TODO:
- [ ] Logic for redirecting after login
- [ ] Logic for redirecting after register

### DIFF
--- a/src/actions/sessionActions.js
+++ b/src/actions/sessionActions.js
@@ -17,7 +17,7 @@ export function unRecognize() {
   }
 }
 
-function callSessionApi(tokens) {
+function callSessionApi(tokens = {}) {
   return (dispatch) => {
     const args = Object.keys(tokens)
       .map((k) => ((tokens[k]) ? `${encodeURIComponent(k)}=${encodeURIComponent(tokens[k])}` : ''))
@@ -104,5 +104,6 @@ export const trackPage = () => {
 
 export const actions = {
   unRecognize,
-  loadSession
+  loadSession,
+  callSessionApi
 }
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -17,6 +17,9 @@ var config = {
     host: "0.0.0.0",
     historyApiFallback: {
       disableDotRule: true
+    },
+    proxy: {
+      "/api": "http://localhost:8000"
     }
   },
   devtool: 'sourcemap',
--- a/src/actions/accountActions.js
+++ b/src/actions/accountActions.js
@@ -1,6 +1,8 @@
 import 'whatwg-fetch'
 
 import Config from '../config.js'
+import { parseServerErrors } from '../lib'
+import { actions as sessionActions } from './sessionActions'
 
 export const actionTypes = {
   REGISTER_SUBMIT: 'REGISTER_SUBMIT',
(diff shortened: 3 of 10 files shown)


src/actions/sessionActions.js
```diff
@@ -17,7 +17,7 @@ export function unRecognize() {
   }
 }
 
-function callSessionApi(tokens) {
+function callSessionApi(tokens = {}) {
```
jparkrr: On login and register, this ends up calling session.json with no tokens, because it reads the cookie that is now expected to be there. Maybe I should call `loadSession`, which also looks for tokens in the url, but they shouldn't exist for login/register.

webpack.config.js
```diff
@@ -17,6 +17,9 @@ var config = {
     host: "0.0.0.0",
     historyApiFallback: {
       disableDotRule: true
+    },
+    proxy: {
+      "/api": "http://localhost:8000"
```
jparkrr: Actually we don't need the proxy right now since I am calling `callSessionApi()`, which doesn't check for cookies https://github.com/MoveOnOrg/mop-frontend/pull/406/files#r179285189, but if I called `sessionActions.loadSession()`, it expects the cookie to be set on this domain. Perhaps there is more of a reason to have a proxy in the future? I'll think about it.

 I feel like this should be safe all the time, since the local data is at /local/api, but idk.

src/actions/sessionActions.js
schuyler1d: I think that's right. If someone explicitly logs in then they shouldn't keep any (different) akid/id token state.

webpack.config.js
schuyler1d: having a proxy will make it easier to have an environment closer to production where we don't need to allow cross-origin requests, so I'm ++ on the proxy.

src/actions/sessionActions.js
```diff
@@ -7,18 +8,24 @@ export const actionTypes = {
   USER_SESSION_FAILURE: 'USER_SESSION_FAILURE'
 }
 
-export function unRecognize() {
+/**
+ * Asynchronously POST to the logout api, mark user as anonymous in the userStore,
+ * and (optionally) push a new route
+ */
+export function unRecognize({ redirect } = {}) {
```
schuyler1d: this should dispatch the UNRECOGNIZE_USER_SESSION right away and not wait until the api call comes back (so for petitions people can immediately start filling out their own info)

src/containers/nav.js
```diff
@@ -47,6 +49,7 @@ class Nav extends React.Component {
         isOpenMobile={this.state.isOpenMobile}
         toggleSection={this.toggleSection}
         openSections={this.state.openSections}
+        onLogout={() => dispatch(sessionActions.unRecognize({ redirect: '/' }))}
```
schuyler1d: ideally this would do this only if 
`/login/do_logout.html` path is marked prodReady

src/routes.js
```diff
@@ -107,6 +107,10 @@ export const routes = store => {
       <Route path='login/register.html' component={LoadableRegister} />
```
schuyler1d: We need a `/login/do_logout.html` (that takes a redirect parameter) here because the perl site will actually still just have a link

src/actions/sessionActions.js
jparkrr: fetch returns a promise and isn't synchronous, so it already works like this

jparkrr:
thanks for reviewing @schuyler1d . everything should be fixed if you look at the last few commits.

Also sorry that the webpack and react updates are in this branch. I had already resolved the conflicts and didn't want to keep doing that. So ideally we can test these things together.

PR #422, opened 2018-04-09
Fix partner logos

### LINKED ISSUE #413: Partner logos not showing on non-mega partner pages
https://petitions.moveon.org/sign/vote-on-national-concealed (old version) shows a partner logo.
https://broom.moveon.org/sign/vote-on-national-concealed (new version) does not, but should. 

https://broom.moveon.org/workingfamilies/sign/say-no-to-unqualified-1 shows a partner logo. Looks like we're just doing that only on mega-partner pages (where the URL includes the partner), but need to also do it on non-mega-partner pages (where the URL doesn't change, but the organization exists in the petition metadata).

jparkrr:
Add a CobrandLogo container that responds to to the organization in the url or loaded petition …

This will display an org logo if there is an orgName specified in the url or the currently loaded petition has a creator.organzation or sponsor.organization.

See this commit for why this wasn't working before https://github.com/MoveOnOrg/mop-frontend/commit/4940e2f3400db9d816b6e18190548f3be62162ae

Fixes #413

### DIFF
--- a/src/containers/cobrand-logo.js
+++ b/src/containers/cobrand-logo.js
@@ -45,7 +45,7 @@ const getCobrandFromPetition = (petition = {}) => {
 }
 
 const mapStateToProps = ({ navStore, petitionStore }, { params }) => {
-  let cobrand = {}
+  let cobrand
 
   // Will be present we are at an org url
   const orgName = params && params.organization
--- a/src/containers/cobrand-logo.js
+++ b/src/containers/cobrand-logo.js
@@ -17,8 +17,9 @@ class CobrandLogo extends React.Component {
   }
 
   render() {
-    const { cobrand } = this.props
-    if (!cobrand) return null
+    const { cobrand, shouldLink } = this.props
+    // Don’t render if org has no logo
+    if (!cobrand || !cobrand.logo_image_url) return null
 
     if (cobrand.browser_url) {
       return <Link to={cobrand.browser_url}>{this.renderLogo()}</Link>
@@ -35,7 +36,6 @@ const getCobrandFromPetition = (petition = {}) => {
   if (!branding) return null
 
   const { organization_logo_image_url, organization, browser_url } = branding
-  if (!organization_logo_image_url) return null
 
   return {
     logo_image_url: organization_logo_image_url, // format of the org api
--- a/src/containers/cobrand-logo.js
+++ b/src/containers/cobrand-logo.js
@@ -17,11 +17,12 @@ class CobrandLogo extends React.Component {
   }
 
   render() {
-    const { cobrand, shouldLink } = this.props
+    const { cobrand, link } = this.props
     // Don’t render if org has no logo
     if (!cobrand || !cobrand.logo_image_url) return null
 
-    if (cobrand.browser_url) {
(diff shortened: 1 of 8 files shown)


src/containers/cobrand-logo.js
```diff
@@ -0,0 +1,71 @@
...
+
+    if (cobrand.browser_url) {
+      return <Link to={cobrand.browser_url}>{this.renderLogo()}</Link>
+    }
+
+    return this.renderLogo()
+  }
+}
+
+const getCobrandFromPetition = (petition = {}) => {
```
jparkrr: Please check my logic against the api here

test/components/cobrand-logo.js
```diff
@@ -0,0 +1,94 @@
...
+        })}
+      />
+    )
+    expect(context.find(Link).prop('to')).to.equal('http://example.com')
+    expect(context.find('img').html()).to.equal(
+      '<img class="org_logo" src="yes.jpg" alt="Move On logo">'
+    )
+  })
+
+  it('sets cobrand using an org url, overriding loaded petition', () => {
```
jparkrr: This is the only completely new test. The rest I converted from reducer tests.

jparkrr:
It's also good to pull this out into a component so we can display the partner logo outside of the nav, as requested for the rebrand.

src/containers/cobrand-logo.js
```diff
@@ -0,0 +1,71 @@
...
+  // grab the sponsor / creator properties from _embedded if it exists
+  const { _embedded: { sponsor, creator } = {} } = petition
+  const branding = sponsor || creator
+  if (!branding) return null
+
+  const { organization_logo_image_url, organization, browser_url } = branding
+  if (!organization_logo_image_url) return null
```
schuyler1d: The logic is here:
https://github.com/MoveOnOrg/mop/blob/main/petitions/serializers.py#L221
A couple issues here:
* something can, in theory, not have a logo, but still have an organization name
* browser_url is available from the organization api, but NOT the petition api.  This might be something we want to change/update on the mop side, since we are abandoning partner pages. (will check with platform).  Currently, the petitions api returns 'orgnaization_url' which is a *local* link for megapartners' petition page (rather than a link to their own homepage)

schuyler1d:
So the specification back from platform is that there should not be any link to partner pages in the header.  I will add `browser_url` to the api, in case we need that inline somewhere.

jparkrr:
Thanks for checking @schuyler1d. Should be fixed now. Note I removed the org home stuff in another pr #425

src/containers/cobrand-logo.js
```diff
@@ -0,0 +1,73 @@
...
+    )
+  }
+
+  render() {
+    const { cobrand, link } = this.props
+    // Don’t render if org has no logo
+    if (!cobrand || !cobrand.logo_image_url) return null
+
+    // `props.link` (defaults to false) will give the logo a link to the partner’s homepage, if defined
+    if (cobrand.browser_url && link) {
```
schuyler1d: so, sorry.  I lied -- after I wrote that comment, we realized that we weren't going to have links at all including to non-existent-going forward partner pages.  so we should just remove this branch and other `browser_url` uses below.

PR #429, opened 2018-04-11
Use env var for bundle analyzer webpack plugin

jparkrr:
I didn't know a better way of doing this besides the refactor

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -9,8 +9,72 @@ var APP_ENTRY = process.env.APP_ENTRY || "main";
 
 var THEME = process.env.THEME || "legacy";
 var THEME_DIR = path.resolve(__dirname, 'src/components/theme-' + THEME);
-const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
+var BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
 
+var envVars = {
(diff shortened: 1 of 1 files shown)


webpack.config.js
```diff
@@ -9,8 +9,72 @@ var APP_ENTRY = process.env.APP_ENTRY || "main";
 
 var THEME = process.env.THEME || "legacy";
 var THEME_DIR = path.resolve(__dirname, 'src/components/theme-' + THEME);
-const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
+var BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
```
schuyler1d: if we load it here, should it still be in devDependencies in package.json?

jparkrr: yep it should. And it's still in there from the react-loadable branch

schuyler1d: I meant more -- should it be moved to just 'dependencies' section since webpack won't build without it?

jparkrr: well, webpack itself plus a lot of things required for build is in devDependencies. The difference doesn't really matter anymore I think. Maybe we should move everything to regular dependences like CRA https://github.com/facebook/create-react-app/pull/2699#issuecomment-312281622

schuyler1d: ok, you've convinced me it doesn't matter, in any case. :+1:

PR #436, opened 2018-04-12
Sign now button fix: Check to see if the form ref is still in the body

### LINKED ISSUE #435: Clicking sign now button does nothing for recognized users
I'm not sure, but I think this is a different issues from #434. You can see it not working here:

https://petitions.moveon.org/2018r/sign/help-test-the-redesign?akid=203343.25362762.1kk7gU

Whereas clicking does work here:

https://petitions.moveon.org/2018r/sign/help-test-the-redesign

jparkrr:
Apparently the ref still existed when the input elements were conditionally hidden.

https://developer.mozilla.org/en-US/docs/Web/API/Node/contains

Fixes #435


jparkrr:
Also I had `country` instead of `address1`

PR #437, opened 2018-04-12
Use scrollchor to scroll from the sign now button

jparkrr:
It has a fixed offset that we can specify, so hopefully it doesn't end up behind the nav.

I haven't tested this much yet.


jparkrr:
Tested on chrome mobile and desktop, as well as iOS safari. Seems to work better everywhere

PR #445, opened 2018-04-16
Revert "Temp fix firefox redirect"

jparkrr:
Reverts MoveOnOrg/mop-frontend#441

https://github.com/MoveOnOrg/mop/pull/348 solves this for realz


schuyler1d:
I kind of like this as a guard -- at least for this week.  Then let's merge this.

Issue #412, opened 2018-04-06
support 404/not found page and a 500/server page


schuyler1d:
Implement 404 and an error page with these graphics:
https://static.moveon.org/giraffe/images/error404.svg
(with the language: "Oops -- looks like what you're looking for isn't here! Email us at help@moveon.org so we can help you find your way.")
and
https://static.moveon.org/giraffe/images/error500.svg
with a message "Uh-oh, something went wrong! Go back and try again please."

The error500 message should display in appropriate moments when a server error occurred (that the user should know about)

Perhaps the language needs to be discussed a little more, and we should be ok with also reporting what exactly failed (it doesn't have to be that exact message for all errors).

schuyler1d:
Also see: https://github.com/MoveOnOrg/giraffe/issues/180 for hopefully design guidance.

schuyler1d:
Final 404 language (the bold text): "Oops -- looks like what you're looking for isn't here! Email us at help@moveon.org so we can help you find your way."
Final 500 top language (bold text): "Uh-oh—it looks like something went wrong. We're sorry about that! Please go back and try again. If you're still having trouble, shoot us a note at help@moveon.org."

This is now unblocked since we have comps.

PR #450, opened 2018-04-16
Add error pages and routes

jparkrr:
Closes partially #412 

I added some docs here https://github.com/MoveOnOrg/mop-frontend/blob/5832a82a8a1c04f66d5071c5fdfb446680cbecf9/docs/EXPLANATION--error-handling.md

- [x] Add legacy error pages
- [x] Add giraffe error pages
- [x] Show error if window.error is set
- [x] Show error on loading petition page with `fetch()` error
  - [x] Server doesn't respond (user could be offline)
  - [x] Server returns error status code
- [ ] Show error on signing petition with `fetch()` error
  - [ ] Server doesn't respond (user could be offline)
  - [ ] Server returns error status code
- [x] Show error on static pages with `fetch()` error
  - [x] Server doesn't respond (user could be offline)
  - [x] Server returns error status code

### DIFF
--- a/src/components/theme-giraffe/error404.js
+++ b/src/components/theme-giraffe/error404.js
@@ -1,6 +1,8 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 
+import Config from '../../config'
+
 export const Error404 = ({ error: { description } }) => (
   <div className='container'>
     <div className='row my-5'>
@@ -18,7 +20,7 @@ export const Error404 = ({ error: { description } }) => (
           )}
         </p>
         <img
-          src='https://static.moveon.org/giraffe/images/error404.svg'
+          src={`${Config.STATIC_ROOT}error404.svg`}
           alt='404 error'
         />
       </div>
--- a/src/components/theme-giraffe/error404.js
+++ b/src/components/theme-giraffe/error404.js
@@ -1,0 +1,30 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+export const Error404 = ({ error: { description } }) => (
+  <div className='container'>
+    <div className='row my-5'>
+      <div className='col-12'>
+        <h1 className='text-align-center'>
+          Oops — looks like what you’re looking for isn’t here!
+        </h1>
+        <p className='text-align-center'>
+          {description}
+          {!description && (
+            <span>
+              Email us at <a href='mailto:help@moveon.org'>help@moveon.org</a>{' '}
+              so we can help you find your way.
+            </span>
+          )}
+        </p>
+        <img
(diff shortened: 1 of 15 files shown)


src/components/theme-giraffe/error404.js
```diff
@@ -0,0 +1,30 @@
...
+    <div className='row my-5'>
+      <div className='col-12'>
+        <h1 className='text-align-center'>
+          Oops — looks like what you’re looking for isn’t here!
+        </h1>
+        <p className='text-align-center'>
+          {description}
+          {!description && (
+            <span>
+              Email us at <a href='mailto:help@moveon.org'>help@moveon.org</a>{' '}
```
schuyler1d: ha ha!  I love this pattern `{' '}` to get around react removing white space.

src/components/theme-giraffe/error404.js
```diff
@@ -0,0 +1,30 @@
...
+              so we can help you find your way.
+            </span>
+          )}
+        </p>
+        <img
+          src='https://static.moveon.org/giraffe/images/error404.svg'
```
schuyler1d: This should use STATIC_ROOT

src/components/theme-giraffe/error500.js
```diff
@@ -0,0 +1,33 @@
...
+              We’re sorry about that! Please go back and try again. If you’re
+              still having trouble, shoot us a note at{' '}
+              <a href='mailto:help@moveon.org'>help@moveon.org</a>.
+            </span>
+          )}
+        </p>
+      </div>
+      <div className='col-12 col-md-5 order-md-first'>
+        <img
+          src='https://static.moveon.org/giraffe/images/error500.svg'
```
schuyler1d: STATIC_ROOT

src/components/theme-giraffe/error500.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+export const Error500 = ({ error: { description } }) => (
```
schuyler1d: Can you put some documentation about the 'error arc' in this file/page -- I feel like I'm going to want to understand how it all shows up or doesn't and this seems like a good place to look for it in the code. -- basically how does SERVER_ERROR get seen, and then the wrapper story of when it displays.

jparkrr:
@schuyler1d Really good point about the '*' route. It would match for every route and we would no longer have the intended behavior of redirecting to `${PROD_URL}${path}` if the component isn't implemented. Implemented routes not marked `prodReady` would have redirected correctly.

It would ALSO have broken where we do appLocation.push('https://....')

So I will remove the fallback route. Thanks!

Issue #459, opened 2018-04-17
Change all the things


sreynen:
Some idea for things to change on the rebranded sign page, but only if they can be done today:

1. Make the form wider (comp coming soon)
2. Give the form a background color (comp coming soon)
3. Auto-focus name field on load
4. Move required fields above non-required fields and hide non-required fields until something is added to required fields
5. Change text above signature count bar to say, e.g. "18,642 signatures. NEW goal - We need 20,000!"
6. Remove "Sign Now" button on desktop, only show on mobile
7. Change "Sign This Petition" button to red, #ea1c24

jparkrr:
here's number 6 https://github.com/MoveOnOrg/giraffe/pull/209

jparkrr:
And number 5 #460

sreynen:
One more change: hide nav completely on thanks page. Looks like maybe the minimal flag on the route just needs to be applied in theme-giraffe.

sreynen:
One more one more: change the "sign now" sticky button to the same red, and change the text on it to  match as well, so both say "sign the petition."

PR #460, opened 2018-04-17
Add signature count copy

jparkrr:
part of #459


jparkrr:
We could also remove the "most recent signers" button thingy

PR #463, opened 2018-04-17
Hide address fields until interaction with form on giraffe

jparkrr:
Part of #459


jparkrr:
I had to disable some tests for this, so we should revisit them if we want to keep this (and there are probably some logic changes we should make)

PR #464, opened 2018-04-17
support email validation the same way django does

schuyler1d:
and make sure we submit unicode from the form


jparkrr:
Unfortunately signatureAddForm doesn't use isValidEmail() yet

https://github.com/MoveOnOrg/mop-frontend/blob/b8b3ec38415a0c8b55afa3209fa17ae5aa329f94/src/containers/signature-add-form.js#L113-L114

PR #466, opened 2018-04-18
support petition-slug suffixes like .fb123 or .html

schuyler1d:
corresponding backend issue:
https://github.com/MoveOnOrg/mop/pull/360

e.g. https://petitions.moveon.org/sign/cursive-writing-is-a.fb50?source=c.fb

### DIFF
--- a/src/containers/cobrand-logo.js
+++ b/src/containers/cobrand-logo.js
@@ -47,7 +47,7 @@ const mapStateToProps = ({ navStore, petitionStore }, { params }) => {
 
   // Will be present if we are viewing a petition
   const petitionSlug = params && params.petition_slug
-  const petition = petitionSlug && petitionStore.petitions[petitionSlug]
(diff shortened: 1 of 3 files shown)


jparkrr:
Won't we also need to update loadPetition in petitionActions?

Is it possible instead to update the route? https://github.com/ReactTraining/react-router/blob/v3/docs/guides/RouteMatching.md#path-syntax

sreynen:
I suggested in the backend PR redirecting these instead of serving them. I suggested that as a future version, thinking it was simpler to merge what we already have. But if this PR is complicated, maybe simpler to just switch to redirects now.

schuyler1d:
@jparkrr how do you propose to do this with a route:
```
:paramName – matches a URL segment up to the next /, ?, or #. The matched string is called a param
```
we need to split before the next `/` and it doesn't accept regexes.

PR #473, opened 2018-04-19
Try identifying for segment on page load

### LINKED ISSUE #471: analytics.identify() not always running with akid on links that have it
We are observing that about half the visits with an akid= link don't run `identify()` in session.

I wonder if this might happen because some people sign it (esp. if they're guest-logged-in with an akid) and click to go to thanks.html before session.json loads for them.

Perhaps we should run analytics.identify() before the session.json call if it's already available in the url (rather than waiting for the session.json result all the time.  If that's the case we would want to get `query.akid.split('.')[1]`

jparkrr:
maybe fixes #471

### DIFF
--- a/src/actions/sessionActions.js
+++ b/src/actions/sessionActions.js
@@ -15,6 +15,14 @@ export function unRecognize() {
   }
 }
 
+let hasIdentified = false
+function identify(id) {
+  if (hasIdentified || !id || !window.analytics || !window.analytics.identify) return
+
+  window.analytics.identify(id) // segment tracking
+  hasIdentified = true
+}
+
 function callSessionApi(tokens) {
   return (dispatch) => {
     const args = Object.keys(tokens)
@@ -32,8 +40,7 @@ function callSessionApi(tokens) {
           tokens
         })
         // segment tracking
-        if (window.analytics && window.analytics.identify
-            && json && json.identifiers && json.identifiers.length) {
+        if (json && json.identifiers && json.identifiers.length) {
           let trackIdentity = null
           json.identifiers.forEach((id) => {
             if (/^actionkit:/.test(id)) {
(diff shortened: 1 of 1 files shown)


src/actions/sessionActions.js
```diff
@@ -15,6 +15,14 @@ export function unRecognize() {
   }
 }
 
+let hasIdentified = false
```
schuyler1d: the one thing we should do is re-identify on login, so maybe add a force=true parameter for that case somehow?

src/actions/sessionActions.js
```diff
@@ -68,11 +75,14 @@ export const loadSession = (location) => {
       || (location && location.query
           && (location.query.akid || location.query.id))) {
     const tokens = {}
-    if (location.query.akid) {
-      tokens.akid = location.query.akid
+    const { id, akid } = location.query
+
+    if (akid) {
+      tokens.akid = akid
+      identify(akid.split('.')[1])
```
schuyler1d: I know this was the rule I gave, but something I might worry about is malformed akid= parameters that break here, so let's test for the split length > 1 before ref'ing it.

src/actions/sessionActions.js
jparkrr: because maybe they could be logged in as another user any time session.json is called (e.g. with a cookie) so they should always be reidentified from that?

schuyler1d: oh, if we don't have login in `main` yet, then i mean the future.  basically if someone completes logging in within the app, then we should redo this.  but if that's not in the current codebase, then it's fine-as-is (but then let's make sure we do this in the auth branch)

jparkrr: I'll add it to https://github.com/MoveOnOrg/mop-frontend/pull/406

PR #492, opened 2018-04-24
updated nav campaign tips link to proxy version

### LINKED ISSUE #491: Update campaigns/campaign tips to point to https://moveon.org/campaign-tips
See https://github.com/MoveOnOrg/mop-frontend/pull/492

sandramchung:
This actually points to the right place right now, but in case it changes in the future e.g. to a wordpress page that is easier to edit, this will use the proxy-configured link instead. Closes #491

### DIFF
--- a/src/components/theme-giraffe/nav.js
+++ b/src/components/theme-giraffe/nav.js
@@ -30,7 +30,7 @@ const Nav = ({
             <NavLink href='https://moveon.org/browse-campaigns'>
               Browse Campaigns
             </NavLink>
-            <NavLink to='https://moveon.org/campaign-tips'>Campaign Tips</NavLink>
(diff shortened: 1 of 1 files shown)


src/components/theme-giraffe/nav.js
```diff
@@ -30,7 +30,7 @@ const Nav = ({
             <NavLink href='https://moveon.org/browse-campaigns'>
               Browse Campaigns
             </NavLink>
-            <NavLink to='/campaign_tips.html'>Campaign Tips</NavLink>
+            <NavLink to='https://moveon.org/campaign-tips'>Campaign Tips</NavLink>
```
jparkrr: We should use `href` instead of `to` for external links now. So we don't send external urls to react router.

PR #501, opened 2018-04-25
Show sign now button on mobile by default

### LINKED ISSUE #500: "sign now" button does not appear until scrolling down
I believe the goal is whenever the form is not in view, then the bottom fixed button 'Sign Now" should be visible.

On Android, and in mobile view on my browsers, if it's not visible initially, then the Sign Now button is NOT visible.  Only when I scroll the form into view and then back out of view is it shown.

cc: @sreynen

jparkrr:
fixes #500 

It's `display: none` on desktop in the css, so we don't have to worry about that.

### DIFF
--- a/src/containers/sign-petition.js
+++ b/src/containers/sign-petition.js
@@ -12,7 +12,8 @@ class SignPetition extends React.Component {
     super(props)
     this.state = {
       deviceSize: null,
-      floatingSignVisible: true
+      floatingSignVisible: false,
+      signFormWasVisible: false
     }
     this.setRef = this.setRef.bind(this)
     this.focusSign = this.focusSign.bind(this)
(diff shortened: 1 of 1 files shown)


jparkrr:
hm. good point.

looks like waypoint loads at first, but we see the button animating off the screen if the sign form is above the fold.

so I'll think about another approach.

jparkrr:
Now everything seems to be working whether or not the sign form is above or below the fold @schuyler1d

schuyler1d:
so the team reviewed this issue and decided that we should do a more careful UI round before making changes.  so @hi0ctane is point on changes here.

@jparkrr sorry for wasting your time.  I think this might still be useful, but we need to wait for some more thought on how to improve signatures by putting some CTA 'above the fold'.

jparkrr:
No worries. I'll close this PR and leave the branch around just in case. We can also see it referenced from #500

PR #506, opened 2018-04-26
Remove all references to petition 'slug' and replace with 'name'

### LINKED ISSUE #485: Remove petition.slug from the petitionStore
We create petition.slug for individual petitions being inserted into the state here https://github.com/MoveOnOrg/mop-frontend/blob/main/src/reducers/index.js#L69

Really we should be using petition.name, because slug is redundant and does not always exist.

We should remove slug so no one is tempted to use it.

See #483 and #484

jparkrr:
fixes #485 

As a follow up to https://github.com/MoveOnOrg/mop-frontend/pull/484, where we made nothing depend on petition.slug, I figured it would be good to remove it from the store.

I also changed all references from "slug" to "name" as, in my opinion, it's better to match what is returned from the API, even if the name isn't perfect.

### DIFF
--- a/docs/EXPLANATION--error-handling.md
+++ b/docs/EXPLANATION--error-handling.md
@@ -20,7 +20,7 @@ We also need to clear this error in the router's `onChange` callback, so if they
 
 When the user enters the app at certain special routes, e.g. the petition sign page, the django server provides the app with some data in `window.preloadObjects`. This saves the time of calling the extra `fetch()`.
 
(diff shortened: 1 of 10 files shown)


jparkrr:
Thanks for the review! Those name suggestions make more sense, and I'm happy to change it if you'd like, but to me it's better to match the API (at least in the `store`) so you can easily understand what is happening if you are looking at an API response. Let me know what you think!

schuyler1d:
i guess the other option is to do a bunch of `let name = petition.name // slug` comments :-/
Nothing is perfect.  I'll say just do what you think is best/clearest.

PR #509, opened 2018-05-01
Upgrade to react 16

jparkrr:
🎆 
- [ ] Concurrently we need to update these urls on the backend template https://github.com/MoveOnOrg/mop/blob/5239c15ef33d753329f1346a91d28bd30b06ee76/petitions/templates/base/base.html#L23-L24


jparkrr:
I haven't heard of anyone running into issues with something breaking in a subtle way since the API remained the same. We can keep it on stage for a few days though.

schuyler1d:
let's merge! :+1:

PR #525, opened 2018-05-15
512 mobile signup test

shakalee14:


src/containers/signature-add-form.js
```diff
@@ -35,7 +37,8 @@ class SignatureAddForm extends React.Component {
     this.validationFunction = {
       email: isValidEmail,
       zip: zip => /(\d\D*){5}/.test(zip),
-      phone: phone => /(\d\D*){10}/.test(phone) // 10-digits
+      phone: phone => /(\d\D*){10}/.test(phone), // 10-digits
+      mobile: mobile => /(\d\D*){10}/.test(mobile)
```
schuyler1d: Ditto here. Let's just have phone

src/containers/signature-add-form.js
```diff
@@ -56,7 +59,8 @@ class SignatureAddForm extends React.Component {
       person: {
         full_name: this.state.name,
         email_addresses: [],
-        postal_addresses: []
+        postal_addresses: [],
+        phone_numbers: []
```
schuyler1d: Unnecessary if just phone

shakalee14:
next step:
- [ ] add ability to turn on test 
- [ ] add random string in url, based on that string, show mobile field etc (odds vs evens)

schuyler1d:
for next steps.  Some suggestions:
* add ability to turn on test: Let's do THIS with environment variable.  Default is off, if AB_TEST_ENABLED.  You'll need to add it to bottom of webpack.dev.js and then also in src/config.js  Then where you test it (see below), you'll import Config and test.  look for other env variables in code to see how that's done (good examples are in actions/petitionActions.js and sessionActions.js)

schuyler1d:
...for changing the url, i think you want to add a parameter -- something like ?abtest_cohort=1 (or 2)
The logic will be something like:
```
if (Config.AB_TEST_ENABLED) {
   if (Math.random() > parseInt(Config.AB_TEST_ENABLED)/100) { //assumes AB_TEST_ENABLED = 10 for 10 percent of users being part of test
      var cohort = (Math.random() > 0.5 ? 1 : 2)
      // set params.cohort here or redirect to window.location + ?&cohort=cohort 
   }
}
```
maybe routes.js or someplace around there on load

schuyler1d:
... then in src/containers/signature-add-form.js mapStateToProps at the bottom, test for `query.cohort` and set displayMobileOptin on the attribute/property accordingly (if it's 1 show it, otherwise don't or whatever)

shakalee14:
not sure why i'm inheriting linting errors from files i didn't work in

src/containers/signature-add-form.js
```diff
@@ -234,11 +236,13 @@ class SignatureAddForm extends React.Component {
         user={user}
         query={query}
         showAddressFields={showAddressFields}
+        showMobileSignup={showMobileSignup}
```
shakalee14: showMobileSignup is whether to display the mobile field in the first place

webpack.config.js
```diff
@@ -23,7 +23,8 @@ var envVars = {
   'STATIC_ROOT': process.env.STATIC_ROOT || '/local/',
   'TRACK_SHARE_URL': process.env.TRACK_SHARE_URL || '',
   'USE_HASH_BROWSING': process.env.USE_HASH_BROWSING || false,
-  'PROD': process.env.PROD
+  'PROD': process.env.PROD,
+  'AB_TEST_ENABLED': process.env.AB_TEST_ENABLED
```
shakalee14: Set it to 10

shakalee14:
@sandramchung i have to run `THEME=giraffe AB_TEST_ENABLED=10 npm run dev` to run this locally

jparkrr:
I'm looking into the linting errors. Seems like something changed in eslint-plugin-react sort-comp in a minor update.

shakalee14:
yea - after removing my node_modules directory, and re-installing, i see the linting errors locally @jparkrr

webpack.config.js
schuyler1d: What does 10 mean?

1. Can you document this in docs/REFERENCE--environment_variables.md
2. Can you add defaults to .travis.yml so it's enabled on staging so we can test there?

schuyler1d: I think you also want an `|| false` here.  webpack is a little weird with empty values.

src/routes.js
```diff
@@ -86,10 +86,30 @@ export const routes = store => {
       store.dispatch(loadOrganization(nextState.params.organization))
     }
   }
+
+  const testFn = () => {
+    const cohort = (Math.random() > 0.5 ? 1 : 2)
+    const currentLocation = window.location
+    const pathName = currentLocation.pathname
+    const check = Config.AB_TEST_ENABLED / 100
```
schuyler1d: AB_TEST_ENABLED will be a string, so let's explicitly convert it to an int here: `parseInt(Config.AB_TEST_ENABLED || 0) / 100`

src/routes.js
```diff
@@ -86,10 +86,30 @@ export const routes = store => {
...
+    if (Config.AB_TEST_ENABLED) {
+      if (Math.random() > parseInt(check, 10)) {
+        // makes sure it only does it on sign pages
+        if (pathName.search('sign') > -1) {
```
schuyler1d: do we also need to check whether `cohort` is already an argument (if it's not we don't want to rerun this)?

src/routes.js
```diff
@@ -86,10 +86,30 @@ export const routes = store => {
       store.dispatch(loadOrganization(nextState.params.organization))
     }
   }
+
+  const testFn = () => {
+    const cohort = (Math.random() > 0.5 ? 1 : 2)
+    const currentLocation = window.location
```
schuyler1d: is there a way to get this from args rather than reference `window`?  for instance, see `orgLoader` in routes.js (this file).

src/routes.js
```diff
@@ -86,10 +86,30 @@ export const routes = store => {
...
+    const pathName = currentLocation.pathname
+    const check = Config.AB_TEST_ENABLED / 100
+    if (Config.AB_TEST_ENABLED) {
+      if (Math.random() > parseInt(check, 10)) {
+        // makes sure it only does it on sign pages
+        if (pathName.search('sign') > -1) {
+          browserHistory.push(`${pathName}?cohort=${cohort}`)
+        } else {
```
schuyler1d: do we need an else{} block here, or will it just fall through to the `return cohort` below?

src/routes.js
```diff
@@ -98,7 +118,8 @@ export const routes = store => {
       {/* Sign pages are popular entry pages, so they get included in the main bundle (not Loadable)
           petitionName is a slugified name, matching the slugified "name" returned by the api.
       */}
-      <Route path='sign/:petitionName' component={Sign} prodReady />
+
+      <Route path='sign/:petitionName' component={Sign} mobileTest={testFn()} prodReady />
```
schuyler1d: i think the way this is run (when route config is instantiated), if people navigate from e.g. the homepage or another part of the react app, then it won't change.  So someone has to land straight to the sign page for this to be triggered.  That might be ok, but then we should document that.

Another way might be to run this onEnter= and then add to the params or something. (see orgLoader)

src/routes.js
shakalee14: i think you only have access to nextState if you attach it to onEnter.

src/components/theme-giraffe/signature-mobile-subscribe.js
```diff
@@ -0,0 +1,34 @@
...
+import { InputBlock } from 'GiraffeUI/input-block'
+import { Link } from 'react-router'
+
+const MobileSubscribe = ({ updateOptIn, updatePhoneNo, showBox }) => (
+  <div>
+    <InputBlock
+      name='phone'
+      label='Mobile'
+      onChange={updatePhoneNo}
+      placeholder='5555555555'
```
shakalee14: Add dashes --> 555-555-5555

src/routes.js
```diff
@@ -86,10 +86,31 @@ export const routes = store => {
...
+    const cohort = (Math.random() > 0.5 ? 1 : 2)
+    const currentLocation = window.location
+    const pathName = currentLocation.pathname
+    const check = parseInt(Config.AB_TEST_ENABLED / 100, 10)
+    if (Config.AB_TEST_ENABLED) {
+      if (Math.random() > check) {
+        // makes sure it only does it on sign pages and
+        // only triggers if you land directly on sign page vs through
+        if (pathName.search('sign') > -1 && pathName.search('cohort') < 0) {
+          browserHistory.push(`${pathName}?cohort=${cohort}`)
```
schuyler1d: This is assuming there are no other query string variables, when often there are.  Let's keep existing query args in here.  @jparkrr is there a good pattern for tihs?

shakalee14: NOTE: need to account for currentLocation.search

src/routes.js
```diff
@@ -86,10 +86,31 @@ export const routes = store => {
       store.dispatch(loadOrganization(nextState.params.organization))
     }
   }
+
+  const testFn = () => {
+    const cohort = (Math.random() > 0.5 ? 1 : 2)
+    const currentLocation = window.location
+    const pathName = currentLocation.pathname
+    const check = parseInt(Config.AB_TEST_ENABLED / 100, 10)
```
schuyler1d: this needs to be `parseInt(Config.AB_TEST_ENABLED, 10) /100`

shakalee14: place inside if statement

src/routes.js
shakalee14: suggestion: `/cohort=/.test(String(currentLocation))`

PR #539, opened 2018-06-01
Remove ONLY_PROD_ROUTES from staging build

jparkrr:
To make it easier to test work in progress


jparkrr:
To me this seems good enough for now. It will be easy enough to flip back on before we deploy anything big to prod. Let me know if you think I should implement a query parameter instead.

PR #544, opened 2018-06-05
Show signing fields if address is required

### LINKED ISSUE #541: Signing fails when logged in user does not have address
When a user is logged in, but session.json does not report having a full local address (i.e. `potential_addresses: [status: "Potential"]` is NOT present), and the petition requires an address, then clicking 'Sign' doesn't do anything.  This is presumably because validation is trying to report that there is no address, but reporting that near the address fields which are hidden when someone is logged in.

Instead, address fields should be shown if someone is not logged in OR (when it is a local petition AND they don't have a potential address on file )

My guess is this was introduced when we hid the additional fields by default and only show them once someone starts typing.

jparkrr:
first, show all the fields after there is a validation error

second, always show all the fields if the user doesn't have an address and the petiton needs one

fixes #541

### DIFF
--- a/test/components/signature-add-form.js
+++ b/test/components/signature-add-form.js
@@ -46,7 +46,7 @@ describe('<SignatureAddForm />', () => {
   const storeAkidAddress = { userStore: { signonId: 123456,
       token: 'akid:fake.123456.bad123',
       given_name: 'Three Stacks',
-      postal_addresses: ['some address']
+      postal_addresses: [{ status: 'Potential' }]
   } }
 
   describe('<SignatureAddForm /> static tests', () => {
--- a/test/components/signature-add-form.js
+++ b/test/components/signature-add-form.js
@@ -21,8 +21,15 @@ describe('<SignatureAddForm />', () => {
   // Below are a set of profiles and user store states that can be re-used
   // for different test conditions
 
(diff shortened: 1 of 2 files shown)


jparkrr:
Hmmm. I guess it should only show the address on needs_full_addresses if you are logged in, and keep them hidden at first if you are logged out (since you already have to type something)

jparkrr:
Hm, @schuyler1d said in #544 
> Instead, address fields should be shown if someone is not logged in OR (when it is a local petition AND they don't have a potential address on file )

test/components/signature-add-form.js
```diff
@@ -36,6 +43,12 @@ describe('<SignatureAddForm />', () => {
     token: 'akid:fake.123456.bad123',
     given_name: 'Three Stacks' } }
 
+  const storeAkidAddress = { userStore: { signonId: 123456,
+      token: 'akid:fake.123456.bad123',
+      given_name: 'Three Stacks',
+      postal_addresses: ['some address']
```
schuyler1d: this is actually an array of dicts, so it doesn't matter, but it might be good to represent what it would actually look like: `postal_addresses: [{status: 'Potential'}]`

PR #547, opened 2018-06-07
Handle empty location in signature

### LINKED ISSUE #545: when city/state are null for signatures, then don't show them
https://petitions.moveon.org/api/v1/petitions/list134655/signatures.json?per_page=10&page=1
(as of reporting) has a user with 'city': null, 'state': null and it's shown as:
"Gabrielle Vxxxx from null"

Instead it should just say 'Gabrielle Vxxxx' without a location.

jparkrr:
fixes #545

### DIFF
--- a/src/containers/signature-list-item.js
+++ b/src/containers/signature-list-item.js
@@ -26,7 +26,7 @@ class SignatureListItem extends React.Component {
     const { user, createdDate, commentId, comments, number } = this.props
     const { city, state } = this.props.user
     const location = ((!city) ? state : `${city}, ${state}`)
-    const fromLocation = ((location === '') ? '' : `from ${location}`)
+    const fromLocation = location && `from ${location}`
 
     return (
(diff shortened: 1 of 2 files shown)


src/containers/signature-list-item.js
```diff
@@ -26,7 +26,7 @@ class SignatureListItem extends React.Component {
     const { user, createdDate, commentId, comments, number } = this.props
     const { city, state } = this.props.user
     const location = ((!city) ? state : `${city}, ${state}`)
-    const fromLocation = ((location === '') ? '' : `from ${location}`)
+    const fromLocation = location && `from ${location}`
```
sandramchung: and now I've learned some magical boolean operator behavior

jparkrr: yeah. It might not be the clearest thing in the world, but it becomes second nature after a while (and a lot of JSX, where it's pretty common)

PR #549, opened 2018-06-07
Adds FormTracker class to lib

shakalee14:


jparkrr:
A small thing, but consider pulling it out to a new file in /lib and putting `export { FormTracker } from './formtracker'` in lib/index.js.

src/lib/form-tracker.js
```diff
@@ -0,0 +1,28 @@
...
+    this.track('form_finished', eventInfo)
+  }
+
+  this.track = function (eventName, eventInfo) {
+    // console.log('eventInfo: '+ JSON.stringify(eventInfo) + ' and eventName: ' + eventName);
+    window.analytics.track({
+      event: eventName,
+      properties: {
+        result: eventName,
+        experiment_id: Config.SEGMENT_TEST_ID,
```
schuyler1d: This might be interesting as a variable but should probably also be something that comes during FormTracker instantiation (i.e. where we do `new FormTracker(...)`) -- because the experiment might be different per-form -- not on a global.

src/lib/form-tracker.js
```diff
@@ -0,0 +1,28 @@
...
+  this.startForm = function (eventInfo) {
+    this.state.formStarted = 1
+    this.track('form_started', eventInfo)
+  }
+
+  this.endForm = function (eventInfo) {
+    this.state.formFinished = 1
+    this.track('form_finished', eventInfo)
+  }
+
```
schuyler1d: I think we want another method something like 
```
this.update = function(info) {
   for (var a in info) { this.state[a] = info[a] } // or the better ES6 way to do this, which I forget -- this.state = { ...this.state, ...info }?
}
```

shakalee14:
Next Steps (after convo) with @schuyler1d 
- [ ] Clean up code (keep state in form tracker vs component, directly call form tracker methods in component)
AND then start keeping track of the following: 
- [ ] Anytime a field changes
- [ ] Anytime a field focuses
- [ ] Anytime theres a validation error
- [ ] Anytime we show more fields (ie volunteer)
- [ ] When someone log ins/we get in new login info

shakalee14:
@schuyler1d - i think this should be reviewed again with the added work

docs/REFERENCE--environment_variables.md
```diff
@@ -17,3 +17,5 @@ THEME                             | Controls whether `import A from "Theme/a"` c
 USE_HASH_BROWSING                 | Whether you want to use react-router HashHistory _Default_: false
 LOCAL_REACT                       | Dev only: a path with react.js and react-dom.js, if you want to develop offline _Default_: Loads react from unpkg.com CDN
 LOCAL_CSS                         | Dev only: a theme-giraffe style.css. E.g. `http://localhost:3000/styles/main.css`, as served by `gulp watch` in the giraffe repo, so you can change css and/or develop offline. _Default_: Loads css from the mop-static-stage s3 bucket
+SEGMENT_TEST_ID                   | Name for segment test to send with event tracking
```
shakalee14: Remove SEGMENT_TEST_ID from env variables -- will be passed through in class instantiation.

src/containers/signature-add-form.js
```diff
@@ -42,6 +42,11 @@ class SignatureAddForm extends React.Component {
     this.submit = this.submit.bind(this)
     this.validationError = this.validationError.bind(this)
     this.updateStateFromValue = this.updateStateFromValue.bind(this)
+    this.formTracker = new FormTracker({ experimentId: 'signMobilePhones1' })
```
shakalee14: Since we have access to props in constructor, can pass through this.props.query.cohort or cohort info in formTracker instance

src/lib/form-tracker.js
```diff
@@ -0,0 +1,164 @@
+import Config from '../config'
+
+export function FormTracker({ experimentId }) {
+  this.state = {
```
shakalee14: can separate out --> 
things that are helpers vs. being sent to segment

src/lib/form-tracker.js
```diff
@@ -0,0 +1,164 @@
...
+
+  this.submitForm = function submitForm(eventInfo) {
+    Object.keys(eventInfo).forEach(key => {
+      this.state[key] = eventInfo[key]
+    })
+    this.track('form_finished')
+  }
+
+  this.formExpandTracker = function formExpandTracker() {
+    this.state.formexpand = 1
```
shakalee14: += 1 
--> count each time

src/containers/signature-add-form.js
```diff
@@ -42,6 +42,11 @@ class SignatureAddForm extends React.Component {
...
+  }
+
+  componentDidUpdate() {
+    this.formTracker.getForm(document, this.props.id)
```
shakalee14: note: find a more reacty way to figure out the form html/structure

shakalee14:
Unfortunately @schuyler1d - our tests do not like how we're passing through `innerRef`

when you run `THEME=giraffe npm run test` you get the following error:
`TypeError: _innerRef is not a function`

TODOs:
- [ ] Fix _innerRef error

shakalee14:
This PR will introduce form tracking events for:
1) When someone starts a sign petition form (so once the form is visible and loaded)
2) When someone submits a valid form

This PR includes work for form drop off.

shakalee14:
@schuyler1d - i think this is ready for another round of review

src/routes.js
```diff
@@ -93,16 +93,16 @@ export const routes = store => {
...
+      const queryString = currentLocation.search
       const check = parseInt(Config.AB_TEST_ENABLED, 10) / 100
 
       if (Math.random() > check) {
         // makes sure it only does it on sign pages and
-        // only triggers if you land directly on sign page vs through
-        if (pathName.search('sign') > -1 && /cohort=/.test(currentLocation.search)) {
+        // only triggers if you land directly on sign page vs through source
+        // checks if there is a cohort already in the querystring so it doesnt add another cohort
+        if (pathName.search('sign') > -1 && queryString.search('cohort') === -1) {
```
shakalee14: we should only push cohort into the browserHistory when /sign/ is present and when cohort is not (like if someone reloads the sign page, we shouldn't add ?cohort=# again)

src/containers/signature-add-form.js
```diff
@@ -43,6 +43,20 @@ class SignatureAddForm extends React.Component {
...
+
+  componentDidMount() {
+    if (this.form) {
+      this.formTracker.setForm(this.form)
+    }
+  }
+
+  componentDidUpdate() {
+    if (!this.state.hideUntilInteract) this.formTracker.setForm(this.form)
+    if (this.state.phone.length === 1 && this.props.query.cohort === '1') {
```
shakalee14: to track when the phone field is expanded to the mobile opt in language

src/containers/signature-add-form.js
```diff
@@ -43,6 +43,20 @@ class SignatureAddForm extends React.Component {
     this.submit = this.submit.bind(this)
     this.validationError = this.validationError.bind(this)
     this.updateStateFromValue = this.updateStateFromValue.bind(this)
+    this.formTracker = new FormTracker({ experiment: 'signMobilePhones1', formvariant: props.id, variationname: (props.query.cohort === '1' ? 'smsSubScribeOption1' : 'current') })
```
schuyler1d: style nit: can we wrap the args to each be on a separate line

src/containers/signature-add-form.js
```diff
@@ -43,6 +43,20 @@ class SignatureAddForm extends React.Component {
...
+  }
+
+  componentDidMount() {
+    if (this.form) {
+      this.formTracker.setForm(this.form)
+    }
+  }
+
+  componentDidUpdate() {
+    if (!this.state.hideUntilInteract) this.formTracker.setForm(this.form)
```
schuyler1d: we should also test for `this.form` here just like in DidMount

src/containers/signature-add-form.js
schuyler1d: I don't get the phone.length == 1. state.phone is the value of the phone -- maybe you're thinking of osdiSignature output where it's an array.

Rather than base it on current state, I would rather test for the differential in props with [prevProps/prevState](https://reactjs.org/docs/react-component.html#componentdidupdate) or when someone clicks something itself.  Ok, I guess I understand -- you're anticipating that someone 'starts typing' and then the length will ==1 they've typed the first digit? -- I would avoid that logic, since it won't nec. trigger if it's filled in by autocomplete rather than someone typing in the field.

src/lib/form-tracker.js
```diff
@@ -0,0 +1,127 @@
...
+      this.state.loginstate = 2
+    }
+  }
+
+  this.setForm = function setForm(ref) {
+    if (ref) {
+      this.form = ref
+      this.state.formLength = ref.elements.length
+
+      if (ref.id === 'mobile-sign' && !!(ref.offsetWidth || ref.offsetHeight || ref.getClientRects().length)) {
```
schuyler1d: 1. let's keep logic/knowledge of what the id can be/is out of FormTracker.  It's fine for the formvariant to be equal to `mobile-sign` and `desktop-sign` (better, even).

Let's make setForm take a second argument which is variantname and then `this.state.formvariant = variantname || ref.id`. Then we will only have one branch instead of one for desktop and mobile.
2. Let's define a method `this.isVisible = () => (!!(ref.offsetWidth || ref.offsetHeight || ref.getClientRects().length))` and use it so it's self-documenting what that means.

src/lib/form-tracker.js
```diff
@@ -0,0 +1,127 @@
...
+        if (this.form.elements[i].name === fieldName) {
+          return i
+        }
+      }
+      return -1
+    }
+    return -1
+  }
+
+  this.updateFormProgress = function updateFormProgress(eventInfo) {
```
schuyler1d: I think we will need/want a conditional in here that
```
if (!this.options.formStarted and this.isVisible()) {
    this.startForm()
}
```
since in our mobile/desktop example, if someone resizes and their values go away, then they've essentially started the other form when they start typing, we should trigger a started again on the mobile version.

src/lib/form-tracker.js
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  this.fieldIndex = function fieldIndex(fieldName) {
+    if (this.form) {
+      for (let i = 0; i < this.form.elements.length; i += 1) {
+        if (this.form.elements[i].name === fieldName) {
+          return i
+        }
+      }
+      return -1
```
schuyler1d: style nit: you don't need this twice -- it will fall through to the lower branch

src/containers/signature-add-form.js
```diff
@@ -204,10 +226,14 @@ class SignatureAddForm extends React.Component {
...
 
     if (this.formIsValid()) {
+      this.formTracker.submitForm({
+        loginstate: (this.props.user.anonymous ? 0 : 1)
+      })
       const osdiSignature = this.getOsdiSignature()
       return dispatch(signAction(osdiSignature, petition, { redirectOnSuccess: true }))
     }
     this.setState({ hideUntilInteract: false }) // show fields so we can show validation error
+    this.formTracker.validationErrorTracker(this.state)
```
schuyler1d: why pass `this.state`?  FormTracker shouldn't nec. know anything about a component's state structure or make assumptions about it.  I think this can be passed without arguments -- we just want a single increment per validation test.

src/lib/form-tracker.js
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+// track field focused
+    if (eventInfo.fieldfocused) {
+      this.state.fieldfocused = Math.max(this.state.fieldfocused, this.fieldIndex(eventInfo.fieldfocused))
+    }
+  }
+
+  this.validationErrorTracker = function validationErrorTracker(eventObj) {
+    Object.keys(eventObj).forEach(key => {
```
schuyler1d: I think you're trying to count the errors -- but i think we just want 'number of times the form errored' -- not how many errors per-try.  So I think we can remove eventObj and just have the whole function be `this.state.validationerror += 1`

shakalee14:
need to do this also:
Can we set a custom tag for the form so we can identify the source for these signups in Revere and therefore target more easily?

src/lib/form-tracker.js
shakalee14: i was counting total amt of errors based on our previous note: 
`count of the number of fields that a validation error occurs before submission`

src/lib/form-tracker.js
shakalee14: fyi i need to pass ref or something into isVisible (based on current logic)

src/containers/signature-add-form.js
```diff
@@ -43,19 +43,26 @@ class SignatureAddForm extends React.Component {
...
+      variationname: (props.query.cohort === '1' ? 'smsSubScribeOption1' : 'current')
+    })
   }
 
   componentDidMount() {
-    if (this.form) {
-      this.formTracker.setForm(this.form)
-    }
+    const ref = this.form
+    if (ref && this.formTracker.isVisible(ref)) this.formTracker.setForm(ref, ref.id)
```
schuyler1d: sorry if I wasn't clear -- I think the isVisible() test should be in FormTracker methods.  But this is fine, too. :man_shrugging:

PR #555, opened 2018-06-11
Submit petition and redirect to thanks page

### LINKED ISSUE #551: After an unregistered user has previewed a new petition, register them and post it to the backend
the last step in petition creation for new users, following up from #339 

So on the petition preview page we need to ensure, if the creator is not already authenticated, we'll create the account before submitting the petition.

We reuse the RegisterForm component that is used on the registration page, and will use its success callback to submit the petition that is being previewed (in the redux store).

jparkrr:
Submits the previewed petition to the api after the user register has succeeded and redirects to /create_finished.html, which is the same thanks page as signing a petition, except `isCreator` is set to true, so some copy and tracking is changed. (I haven't fully verified everything)

For now I just made it so any server error from the create petition endpoint will just show the 500 error. Should we consider something like taking them to the revise page if its a 400 error? I'm not sure what kind of errors would cause the petition to be rejected. (I'll try looking into this)

I'll try to add some tests, but this seems to be working well with mop running locally and `API_URI=/api/v1 API_WRITABLE=1` in the env :)

closes #551


src/actions/createPetitionActions.js
```diff
@@ -18,6 +23,44 @@ export function previewSubmit({ title, summary, description, target }) {
...
 }
 
+export function submit() {
+  return (dispatch, getState) => {
+    dispatch({
+      type: actionTypes.CREATE_PETITION_REQUEST
+    })
+    const { petitionCreateStore: p } = getState()
+    return (
+      fetch(`${Config.API_URI}/user/petitions.json`, {
```
schuyler1d: I think we should condition this on the API_WRITABLE config var, for easier testing.
Also, should we set the Content-Type?  The backend, to my knowledge won't do anything specific with it, but it might be good practice.  application/json? or if you want to go nuts: application/hal+json

src/containers/create-finished.js
```diff
@@ -0,0 +1,35 @@
...
+        <Thanks isCreator petition={this.props.petition} />
+      </div>
+    )
+  }
+}
+
+function mapStateToProps({ petitionCreateStore }) {
+  return {
+    hasSubmittedPetition: !!(petitionCreateStore && petitionCreateStore.submitted),
+    petition: petitionCreateStore.petition
```
schuyler1d: What if they create more than one in a session?  is this always the last petition created?  I guess that seems fine -- it actually might be useful for the dashboard view when we return to that.

jparkrr: Yes, the last one. And it's explicitly cleared out when they preview a new petition they are creating.

src/actions/createPetitionActions.js
jparkrr: Good point. Also made me realize I wasn't including cookies. So I wonder how the it was working. hmmmm looking into this

jparkrr: ok so somehow it's already including cookies for me, but I'll change it to be right anyway

jparkrr:
@schuyler1d I made it so API_WRITABLE=false (the default) makes no requests and succeeds!

schuyler1d:
merge at will

PR #561, opened 2018-06-14
Have travis cache node_modules

jparkrr:
For faster builds.

I don't think (?) it will mess anything up, because it will still run `npm install`, just with the `node_modules` folder already the same as last build.

https://blog.travis-ci.com/2013-12-05-speed-up-your-builds-cache-your-dependencies


jparkrr:
woo. down to 1 min 45 sec

PR #569, opened 2018-06-25
Attempts to correct track api call

shakalee14:


shakalee14:
With these changes, I can now see the following`form_started` and `form_finished` with the correct fields in the segment schema interface

PR #571, opened 2018-06-29
On signing, validate that the name is not an email address

### LINKED ISSUE #560: Safari seems to autocomplete email in name field
Autocomplete seems to have a number of issues.  Based on user data, it seems safari mobile autocompletes the email address in the name field.

I've seen the same happen in firefox in the petition creation registration form.

Perhaps validation should disallow '@' in the name field?

jparkrr:
hopefully closes #560, but doesn't really get to the root of the problem as to why people are filling "name" with emails. I tried to reproduce in mobile safari, but couldn't.


jparkrr:
I figured this captured the intent better

PR #577, opened 2018-07-17
WIP: Create petition tracking (theme-legacy)

jparkrr:
Here I'm starting to use FormTracker to track the theme-legacy create-petition.

It becomes a bit complicated because this is a multistep form, each with separate routes. I solved this by adding a few methods to the tracker to save and load the state from the previous step (for simplicity, handled manually in the component). So it won't track `form_finished` until the petition is actually created.

It seems like this might be an okay way of comparing between the petition creation methods: which one has a better success rate (`total completed`/`total started`), but I'm sure there are complexities I'm not thinking of.

I also utilize the tracker state var `sectionadvanced`, so we'll theoretically be able to know where they were on the multistep form with the combination of `fieldchanged` (where the index resets for the next step) and `sectionadvance` (which will be 0 for the first step and 1 for the second) (really only relevant once bounce tracking is working). But there might be a better way of doing this.

I was also playing around with the idea of creating components like `TrackedForm` and `TrackedInput` to hide all of this complexity and call into formTracker automatically, but not sure if that is where I should be spending my time.

Here's what would get tracked
```
{  
   "result":"completed",
   "variationname":"",
   "formvariant":"",
   "experiment":"",
   "loginstate":0,
   "validationerror":0,
   "formexpand":0,
   "sectionadvanced":1,
   "fieldfocused":-1,
   "fieldchanged":5,
   "lastfieldchanged":"passwordConfirm",
   "lastfieldfocused":"passwordConfirm",
   "formLength":7
}
```

TODO:
- [ ] loginstate tracking
- [ ] full form tracking for logged in state?
- [ ] track the other creation forms when they are ready, and set up an experiment

Let me know if you have any thoughts.

### DIFF
--- a/docs/EXPLANATION--form-tracking.md
+++ b/docs/EXPLANATION--form-tracking.md
@@ -1,0 +1,120 @@
+<a name="FormTracker"></a>
+
+## FormTracker([params])
+A FormTracker instance gives the ability to keep analytics data for forms all in
+the same place, and provides methods for manipulating this analytics data (stored as
+internal state) as well as tracking this data to an analytics provider.
+
+Regenerate the markdown file for these docs with:
+`npx jsdoc-to-markdown src/lib/form-tracker.js > docs/EXPLANATION--form-tracking.md`
+(only edit these docs in form-tracker.js)
+
+**Kind**: global function  
+
+| Param | Type | Description |
+| --- | --- | --- |
+| [params] | <code>object</code> | an object of params (see below) |
+| [params.formvariant] | <code>string</code> | e.g. 'mobile' or 'desktop' |
+| [params.experiment] | <code>string</code> | name of the experiment being run |
+| [params.variationname] | <code>string</code> | the experiment group the user is in |
+
+
+* [FormTracker([params])](#FormTracker)
+    * [.submitForm([eventInfo])](#FormTracker+submitForm)
+    * [.formExpandTracker()](#FormTracker+formExpandTracker)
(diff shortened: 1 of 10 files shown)


docs/EXPLANATION--form-tracking.md
```diff
@@ -0,0 +1,120 @@
+<a name="FormTracker"></a>
```
jparkrr: This file is autogenerated by the jsdocs I added to [form-tracker.js](https://github.com/MoveOnOrg/mop-frontend/blob/f303f7caa3e02359c682aa9a12e3b8f1b0b93be0/src/lib/form-tracker.js#L9). Let me know if you think it's not needed. Might be kinda annoying to keep it up-to-date (running cli tool)

src/lib/form-tracker.js
```diff
@@ -66,6 +100,12 @@ export function FormTracker({ experiment = '', formvariant = '', variationname =
     }
   }
 
+  /**
+   * @private
+   * Finds the index of the field in the saved form, for use in updateFormProgress()
```
jparkrr: So this is the thing that would need to be overridden for use in something like the conversation interface. I'm wondering if that motivates a different design for the tracked values. We could have something like a subclass for each type of form we want to track, so it could have special values like `title: true/false`, `description: true/false`. But maybe we don't need data of that granularity? And I guess especially for the conversation interface, they must do it sequentially.

jparkrr: So nevermind my suggestion, but let me know if you think just having the max input index and fieldname is enough to do the required analysis.

PR #582, opened 2018-07-31
Allows someone to add sms in qs in order to add random cohort

shakalee14:
In order to test on FB, or per petition, we need the ability to trigger it per petition vs per all petitions. I suggest adding 'sms' in the querystring (could even appear as `source=sms`) in order to enable random cohort rendering.


shakalee14:
@schuyler1d - updated to use test regex instead of js search method

PR #583, opened 2018-11-30
only us territories to restrict our legal exposure to USA

schuyler1d:


schuyler1d:
I think we have them in there because sometimes people choose it from country. They are also available in the 'state' field (though that's not technically correct either, which is why it can be called 'region' but that just confuses everyone, then....).  I suppose we could remove the country field altogether and then do the similar giraffe PR where we have a message that says "US and US territories only"

PR #586, opened 2018-12-06
585 perm mobile field

### LINKED ISSUE #585: Add Permanent Mobile Field and Opt In Box for Petition Signers
Related to testing involved with : https://github.com/MoveOnOrg/mop-frontend/issues/512

Need to add the mobile field but remove testing part (cohort etc)

@schuyler1d - is it worth keeping the cohort logic somewhere for future tests (i could put it into a doc) or i could keep the testFn(), add comments https://github.com/MoveOnOrg/mop-frontend/blob/main/src/routes.js#L92 and then just take all the testing pieces out.

shakalee14:
Resolves #585

### DIFF
--- a/src/routes.js
+++ b/src/routes.js
@@ -89,6 +89,8 @@ export const routes = store => {
     }
   }
 
+  // can be used to divide users into random cohorts
+  // previously used for mobile field test : https://github.com/MoveOnOrg/mop-frontend/issues/512
   const testFn = () => {
     const cohort = (Math.random() > 0.5 ? 1 : 2)
     if (Config.AB_TEST_ENABLED) {
@@ -124,7 +126,7 @@ export const routes = store => {
           petitionName is a slugified name, matching the slugified "name" returned by the api.
       */}
 
-      <Route path='sign/:petitionName' component={Sign} mobileTest={testFn()} prodReady />
+      <Route path='sign/:petitionName' component={Sign} prodReady />
       <Route path=':organization/sign/:petitionName' component={Sign} onEnter={orgLoader} prodReady />
 
       <Route path='pac/' component={LoadablePacHome} prodReady />
--- a/src/routes.js
+++ b/src/routes.js
@@ -89,8 +89,14 @@ export const routes = store => {
     }
   }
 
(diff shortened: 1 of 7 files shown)


src/routes.js
```diff
@@ -101,14 +109,15 @@ export const routes = store => {
         // makes sure it only does it on sign pages and
         // only triggers if you land directly on sign page vs through source
         // checks if there is a cohort already in the querystring so it doesnt add another cohort
-        if (/sign/.test(pathName) && /sms/.test(queryString) && !/cohort/.test(queryString)) {
+        if (/sign/.test(pathName) && !/cohort/.test(queryString)) {
```
schuyler1d: I think this would make all /sign/ pages redirect, but instead I think we don't want any pages to redirect. Though I suppose the point is this testFn isn't run at all anymore.  So this is all academic with how we keep this logic in the code.

Let's rename testFn => routeCohortSplitter or something and then document it with "Add `<Route path='....' cohort={routeCohortSplitter()} /> to enable cohorts for a route below.`
Then here, let's make the test abstracted so it's an argument, so routeCohortSplitter(cohortQueryText) =>....
`const addCohortQuery = (new Regex(cohortQueryText)).test(queryString)`
and then add `&& addCohortQuery &&` 'back' into the if-line.

test/components/signature-add-form.js
```diff
@@ -336,70 +336,33 @@ describe('<SignatureAddForm />', () => {
...
+      const context = mount(<SignatureAddForm {...propsProfileBase} store={store} />)
 
       context.find('input[name="name"]').simulate('change', { target: { value: 'Jane Doe' } })
       expect(context.find('input[name="phone"]').length).to.equal(1)
       expect(context.find('input[name="mobile_optin"]').length).to.equal(0)
     })
 
     it('shows an optin check box when mobile field has a value', () => {
-      if (process.env.THEME !== 'giraffe' && process.env.AB_TEST_ENABLED !== 10) return
+      if (process.env.THEME !== 'giraffe') return
```
schuyler1d: style nit: please add braces.

PR #591, opened 2018-12-14
Suggest common domains with typos

schuyler1d:
This is to reduce errors when people enter their email address and sometimes have a typo in the domain name.

This adds immediate feedback for the name and email fields (rather than waiting for submitting) and in the case of an 'off-by-1' difference for common domains, we suggest the more common domain in red (but don't block submission)

petitionemailsuggest


src/lib/index.js
```diff
@@ -116,13 +116,33 @@ try { RegExp('\\w', 'iu') } catch (err) { supportsUnicode = false }
 const userRegex = RegExp('^[-!#$%&\'*+/=?^_`{}|~\\w]+(\\.[-!#$%&\'*+/=?^_`{}|~\\w]+)*$', supportsUnicode ? 'iu' : 'i')
 const quotedUserRegex = /^"([\001-\010\013\014\016-\037!#-[\]-\177]|\\[\001-\011\013\014\016-\177])*"$/i
 const domainRegex = /^((?:[A-Z0-9](?:[A-Z0-9-]{0,61}[A-Z0-9])?\.)+)(?:[A-Z0-9-]{1,62}[A-Z0-9])$/i
+const commonDomains = ['aol.com', 'comcast.net', 'gmail.com', 'hotmail.com', 'verizon.net', 'yahoo.com']
```
shakalee14: how did you come up with this list? i'd suggest adding icloud associated domains too just based on the quick google searches so `icloud.com`, `me.com`, etc.

schuyler1d: I added icloud.com but I think me.com is too short for it to work -- i.e. a one-letter change wouldn't make the 80% threshold and weakening that will make the warning too noisy (false-positives).  Hopefully me.com is short enough it gets fewer typos!

PR #598, opened 2019-03-26
WIP: Cohort Tracker for Segment for WhatsApp

shakalee14:
Adds cohort tracker to tell segment when whatsapp button is displayed


shakalee14:
tracking event being sent to segment looks like this -->
`{"experiment":"whatsAppShare","variationname":"cohort1","user_sighash":"xxxx", user_signonid: xxxx}` 

user_sighash, usersignonid, experiment and variationnname will appear as columns in the segment table

PR #601, opened 2019-03-27
Broom-main

shakalee14:
We want to run a test for several weeks without petition senders needing to include something in the url. This adds the ability to set AB_TEST_ENABLED=50 which then gives folks a 50% chance of seeing the whatsapp button upon signing.


src/actions/sessionActions.js
```diff
@@ -114,9 +114,9 @@ export const loadCohort = location => {
   const cohortString = (location && location.query && location.query.cohort)
   if (cohortString) {
     cohort = Number(cohortString) || 0
-  } else if (location && location.query && 'ab' in location.query) {
+  } else if (Config.AB_TEST_ENABLED === 1 || (location && location.query && 'ab' in location.query)) {
```
schuyler1d: first Config.AB_TEST_ENABLED is probably a string, so it will never === 1.

I think we just want this to be plain test of existence for Config.AB_TEST_ENABLED and then remove the `&& Config.AB_TEST_ENABLED !== 1` below.

When we deploy, we will set `AB_TEST_ENABLED=50` which will make the probability .5 but explicitly -- and that will be good -- that way, if we can see it working/successful, we can just bump it up to AB_TEST_ENABLED=80 or whatever.

shakalee14:
Using this PR for testing but separated out the testing changes into a separate pr (without the bigger wapp button).

PR #603, opened 2019-04-02
596 small link and big button

shakalee14:
Connects to giraffe pr: 
https://github.com/MoveOnOrg/giraffe/pull/258


shakalee14:
Desktop:

Screen Shot 2019-04-02 at 4 28 20 PM

Mobile:

Screen Shot 2019-04-02 at 4 28 32 PM

src/components/theme-giraffe/whatsapp-button.js
```diff
@@ -2,12 +2,12 @@ import React from 'react'
...
 
 import { withWhatsApp } from '../../containers/hoc-whatsapp'
-import WhatsAppSvg from 'GiraffeUI/svgs/whatsapp.svg'
+import WhatsAppWhiteSvg from 'GiraffeUI/svgs/whatsappwhite.svg'
 
 const WhatsAppButton = ({ onClick }) => (
-  <a className='mo-btn petition-thanks__link' onClick={onClick}>
-    <WhatsAppSvg />
-    WhatsApp
+  <a className='petition-thanks__cta wamobile' onClick={onClick}>
```
schuyler1d: where are classes `wamobile` and `wadesktop` used?  If not anywhere, let's remove. -- or the idea is so you can inspect in debugging?  If so, maybe let's name the class `debug--wamobile` or something.

src/giraffe-ui/svgs/whatsappwhite.svg
```diff
@@ -0,0 +1,21 @@
+<svg version="1.1" id="whatsappwhite" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
```
schuyler1d: Can we add this svg file into giraffe repo as well so we can use elsewhere.  Ideally, we would also add a what's app example button here:
https://static.moveon.org/giraffe/styleguide-social.html
(giraffe/src/templates/pages/styleguide/social.twig)

src/components/theme-giraffe/whatsapp-button.js
shakalee14: @schuyler1d 
they are used here - https://github.com/MoveOnOrg/giraffe/pull/258/files - since one is displayed only on mobile and the other is displayed only on desktop. i can try to do it using nth child logic and remove the classes after the test has completed (it'll be hard to choose the child since the number elements will change based on cohorts).

PR #606, opened 2019-04-15
WIP: 595 messenger share

shakalee14:
Test to only show messenger button on mobile.

*Note*:
This pr is dependent on changes in backend and css. It is also not mergeable until WhatsApp testing is finished.

### DIFF
--- a/src/containers/thanks.js
+++ b/src/containers/thanks.js
@@ -9,6 +9,7 @@ import ThanksComponent from 'Theme/thanks'
 import TwitterButton from 'Theme/twitter-button'
 import FacebookButton from 'Theme/facebook-button'
 import WhatsAppButton from 'GiraffeTheme/whatsapp-button'
+import MessengerButton from 'GiraffeTheme/messenger-button'
 import MailButton from 'Theme/mail-button'
 import CopyPaste from 'Theme/copy-paste'
 import RawLink from 'Theme/raw-link'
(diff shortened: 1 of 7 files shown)


src/giraffe-ui/svgs/messenger.svg
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
```
schuyler1d: Can we get a giraffe PR with this, as well?

webpack.config.js
```diff
@@ -27,7 +27,8 @@ var envVars = {
   'TRACK_SHARE_URL': process.env.TRACK_SHARE_URL || '',
   'USE_HASH_BROWSING': process.env.USE_HASH_BROWSING || false,
   'PROD': process.env.PROD,
-  'AB_TEST_ENABLED': process.env.AB_TEST_ENABLED || false
+  'AB_TEST_ENABLED': process.env.AB_TEST_ENABLED || false,
+  'MESSENGER_APP_ID': process.env.MESSENGER_APP_ID,
```
schuyler1d: Adding this will need to get an update in our internal wiki documentation on MOP frontend production deploy.

src/containers/thanks.js
```diff
@@ -108,6 +110,17 @@ class Thanks extends React.Component {
       : '')
   }
 
+  renderMessenger() {
+    return (
```
schuyler1d: Can we condition this on Config.MESSENGER_APP_ID existing?  If it doesn't then we shouldn't render anything, right?

PR #608, opened 2019-04-17
WIP: WhatsApp Test Finishing Prep Work

shakalee14:
In the event that WA is implemented
- [ ] potentially refactor hoc- (?)


shakalee14:
on pause until we figure out results from test.

PR #610, opened 2019-04-26
609 branded test

shakalee14:
Test slated to happen after messenger test with branded colors for Facebook, WhatsApp and Messenger buttons. Twitter button branded color is similar to MO's branded color so not necessary to change that.


shakalee14:
Mobile Comp:
Screen Shot 2019-04-26 at 11 33 35 AM

PR #611, opened 2019-05-02
595 messenger share wo whatsapp

shakalee14:
I need to take the whatsapp button out to test the messenger button separately; some notes below:
- This contains a lot of the logic in #606 
- This doesn't completely rip out whatsapp work because we're still making a decision about whether to keep testing the button. There's a scenario where we could still test messenger and whatsapp.
- This changes the experiment to `messenger2` and populates the messenger table for testing
- This button only appears on mobile when cohort=1

Screen Shot on Mobile:
Screen Shot 2019-05-02 at 11 27 31 AM

### DIFF
--- a/src/containers/hoc-messenger.js
+++ b/src/containers/hoc-messenger.js
@@ -1,0 +1,64 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { petitionShortCode } from '../lib'
+import Config from '../config'
+
+
+function getMobileMessengerLink(encodedValue) {
+  // testing messenger link on mobile only first
+ return `fb-messenger://share?link=${encodedValue}&app_id=${encodeURIComponent(Config.MESSENGER_APP_ID)}`
+}
+
+export function withMessenger(WrappedComponent) {
+  class Messenger extends React.Component {
+    constructor(props) {
+      super(props)
+      this.getShareLink = this.getShareLink.bind(this)
+      this.shareMessenger = this.shareMessenger.bind(this)
+    }
+
+    getShareLink() {
+      const { shortLinkMode, shortLinkArgs } = this.props
+      const messengerShareLink = petitionShortCode(
+        shortLinkMode,
+        ...shortLinkArgs
+      )
+      return messengerShareLink
+    }
+
+    shareMessenger() {
+      /* open this window in case user does not have messenger installed */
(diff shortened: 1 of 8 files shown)


src/containers/thanks.js
```diff
@@ -146,6 +148,19 @@ class Thanks extends React.Component {
     )
   }
 
+  renderMessenger() {
+    const isMobile = /iPhone/.test(navigator.userAgent) || /Android/.test(navigator.userAgent)
+    return (isMobile && this.state.messenger && Config.MESSENGER_APP_ID ?
+      <MessengerButton
+        petition={this.props.petition}
+        shortLinkMode={this.props.isCreator ? 'd' : 'a'}
```
schuyler1d: This requires merge and deploy of this first:
https://github.com/MoveOnOrg/mop/pull/469

src/containers/hoc-messenger.js
```diff
@@ -0,0 +1,64 @@
...
+        shortLinkMode,
+        ...shortLinkArgs
+      )
+      return messengerShareLink
+    }
+
+    shareMessenger() {
+      /* open this window in case user does not have messenger installed */
+      // takes them to moveon page but doesnt include info on petition
+      const encodedValue = encodeURIComponent(this.getShareLink())
```
schuyler1d: This is a nit, but mostly you would put the `encodeURIComponent()` in getMobileMessengerLink, so this part would read:
`const shareLink = getMobileMessengerLink(this.getShareLink())`
and then you would encode the value in getMobileMessengerLink instead of passing it in.

src/containers/hoc-messenger.js
```diff
@@ -0,0 +1,64 @@
...
+      const { shortLinkMode, shortLinkArgs } = this.props
+      const messengerShareLink = petitionShortCode(
+        shortLinkMode,
+        ...shortLinkArgs
+      )
+      return messengerShareLink
+    }
+
+    shareMessenger() {
+      /* open this window in case user does not have messenger installed */
```
schuyler1d: Can you add some documentation for how this works? 
I think the way this works is that your first `window.open()` call will fail if messenger is not available/installed, and so the current page will be unloaded before the second triggers (3 seconds later) -- is that correct?  Does the second timeout get called even if the first call does work?

if the first is called successfully, does recordShare and afterShare get triggered?

PR #626, opened 2019-09-24
send a segment event for shareclick events so we can map shareclicks …

schuyler1d:
…with useragent, etc.

This should allow us to map shareclick info to segment data via segment's anonymous id through r_hash or user_id depending on which one is matchable

### DIFF
--- a/src/actions/petitionActions.js
+++ b/src/actions/petitionActions.js
@@ -293,6 +293,14 @@ export const recordShareClick = (petition, tracking, medium, source, user) => {
   if (user && user.signonId) params.user_id = user.signonId
   if (tracking && tracking.r_hash) params.r_hash = tracking.r_hash
 
+  if (window.analytics) {
+    window.analytics.track('mopshareclick', {
+      ...params,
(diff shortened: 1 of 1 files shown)


src/actions/petitionActions.js
```diff
@@ -293,6 +293,14 @@ export const recordShareClick = (petition, tracking, medium, source, user) => {
   if (user && user.signonId) params.user_id = user.signonId
   if (tracking && tracking.r_hash) params.r_hash = tracking.r_hash
 
+  if (window.analytics) {
+    window.analytics.track('mopshareclick', {
```
sandramchung: This works instrumented as an event. I'm taking the opportunity to call attention to segment's `alias` method https://segment.com/docs/spec/alias/ which I learned about from Jacinda as a way to tie together different identifiers for the same person. So you could alias r_hash to the user's anonymous ID that they already have from segment.

schuyler1d: @sandramchung updated with a commit to alias() anon => r_hash

